PR #1, opened 2020-05-05
Type alias for codec and minor improvments

Liamsi:
Introduce a `Codec` type and go fmt (hence review with "Hide whitespace changes" in the PR diff view: https://github.com/LazyLedger/rsmt2d/pull/1/files?diff=unified&w=1)

### DIFF
--- a/datasquare.go
+++ b/datasquare.go
@@ -189,11 +189,11 @@ func (ds *dataSquare) ColumnRoots() [][]byte {
 	return ds.columnRoots
 }
 
-func (ds *dataSquare) computeRowProof(x uint, y uint) ([]byte, [][]byte, uint, uint) {
+func (ds *dataSquare) computeRowProof(x uint, y uint) ([]byte, [][]byte, uint, uint, error) {
 	tree := merkletree.New(ds.hasher)
 	err := tree.SetIndex(uint64(y))
 	if err != nil {
-		panic(err)
+		return nil, nil, 0, 0, err
 	}
 	data := ds.Row(x)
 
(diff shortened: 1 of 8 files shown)


datasquare.go
```diff
@@ -1,242 +1,248 @@
...
 }
 
 func (ds *dataSquare) computeRowProof(x uint, y uint) ([]byte, [][]byte, uint, uint) {
-    tree := merkletree.New(ds.hasher)
-    tree.SetIndex(uint64(y))
-    data := ds.Row(x)
+	tree := merkletree.New(ds.hasher)
+	err := tree.SetIndex(uint64(y))
+	if err != nil {
+		panic(err)
```
musalbas, line 196: Shouldn't we pass the error upstream if an invalid tree index is supplied?

Liamsi: Yes, absolutely! Good catch. All other places were in tests where it was OK to panic I guess. Will update in a bit.

datasquare.go
```diff
@@ -1,242 +1,248 @@
...
 
 func (ds *dataSquare) computeColumnProof(x uint, y uint) ([]byte, [][]byte, uint, uint) {
-    tree := merkletree.New(ds.hasher)
-    tree.SetIndex(uint64(x))
-    data := ds.Column(y)
+	tree := merkletree.New(ds.hasher)
+	err := tree.SetIndex(uint64(x))
+	if err != nil {
+		panic(err)
+	}
```
adlerjohn, line 213: Here too?

Liamsi: done in 0a13699

Liamsi:
Added some more error handling gotchas in dee82e7

PR #7, opened 2020-09-22
fix incorrect error message about number of chunks required

### LINKED ISSUE #6: Slightly misleading error message
https://github.com/lazyledger/rsmt2d/blob/ea94438fa457081948e455b6b6f2210458a9d699/datasquare.go#L21-L24

The error message suggests that the input length must be powers of two (2^n) while the method accepts all inputs which are square numbers (n^2). The latter makes more sense in this context.

musalbas:
closes #6


liamsi (approved):
👍

Issue #8, opened 2020-10-29
Allow user to choose underlying Merkle tree implementation


musalbas:
See https://github.com/lazyledger/lazyledger-core/pull/83#discussion_r513560314

liamsi:
This might imply that we need to change the API of the NMT to take in bytes instead of that `namespace.Data` interface: https://github.com/lazyledger/nmt/blob/1a5d9124ad10e40d0d743d7695e1983992089a2f/nmt.go#L236

liamsi:
resolved in #9

PR #9, opened 2020-10-30
enable the user to choose a concrete tree implementation #8

liamsi:
ref #8


tree.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+var DefaultTreeCreator = &defaultTreeCreator{}
+
+type Tree interface {
+	Push(data []byte)
+	// TODO(ismail): is this general enough?
+	Prove(idx int) (merkleRoot []byte, proofSet [][]byte, proofIndex uint64, numLeaves uint64)
+	Root() []byte
+}
```
liamsi, line 17: Please don't merge yet. Trying to distill a minimal interface. This one is basically nebolouslabs' API (and these are the methods used).

tree.go
```diff
@@ -0,0 +1,62 @@
...
+
+type TreeCreator interface {
```
musalbas, line 19: I think TreeFactory would be a more conventional name?

liamsi: Yeah, maybe. The reason I didn't call this factory is that it currently is meant to create and return one concrete implementation of a `Tree` (`New` doesn't take in any params). Kind of a boring factory  (at least currently).

tree.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/NebulousLabs/merkletree"
+)
+
+var DefaultTreeCreator = &defaultTreeCreator{}
+
+type Tree interface {
+	Push(data []byte)
+	// TODO(ismail): is this general enough?
+	Prove(idx int) (merkleRoot []byte, proofSet [][]byte, proofIndex uint64, numLeaves uint64)
```
liamsi, line 15: Compare: https://github.com/lazyledger/nmt/blob/1a5d9124ad10e40d0d743d7695e1983992089a2f/nmt.go#L110
https://github.com/lazyledger/nmt/blob/1a5d9124ad10e40d0d743d7695e1983992089a2f/proof.go#L19

liamsi:
ref: https://github.com/evan-forbes/rsmt2d/commit/2ea46c3fb8f76f85dc1d60c72821e420ddd2fb6a

@evan-forbes started an alternative approach which abstracts away the Merkle tree altogether in favour of a "Prover" - it combines the aggregation with generating the proof. Hence a tree creator / factory becomes obsolete. Need to think a bit further about this and if there are any undesired side-effects but so far I like Evan's approach better as it would mean a single simple interface.

evan-forbes:
In my second attempt, with guidance from @liamsi, I've attempted to further isolate data erasure and commitment generation. Instead of configuring, generating, and keeping track of row and column roots using a datasquare, this experiment delgates that functionality to the VectorCommitmentProver.
```go
// VectorCommitmentProver describes the expected methods for issuing and proving
// vector commitments over portions of a data square
type VectorCommitmentProver interface {
	// Commitment provides the commitment root for the selected row or column.
	// Returns a commitment to a zeroed slice if the index is too high.
	Commitment(a Axis, idx uint) []byte
	// Prove issues a proof that can verify the inclusion of a selected row or column.
	Prove(a Axis, idx uint) (Proof, error)
}
```
The VectorCommitmentProver is meant to work with the following types. Proof was taken from this PR draft.
```go
// Axis indicates to the VectorCommitmentProver whether to use columns or rows
type Axis bool

const (
	// Column indicates use of the x axis of a data square
	Column Axis = true
	// Row indicates use of the y axis of a data square
	Row Axis = false
)

// Proof describes the data needed to verify a rsmt2d compatible merkle tree
type Proof struct {
	Root   []byte
	Set    [][]byte
	Index  uint64
	Leaves uint64
}
```
Making VCP implementations will likely consist of wrapping an extended data square. For example, if we want to make a name spaced merkle tree version, we configure the types of nmts we wish to create, and define the expected methods (leaving Prove out for brevity)
```go
// VCP uses configured name spaced merkle trees to adhere to the
// rsmt2d.VectorCommiterProver interface
type VCP struct {
	setters   []nmt.Option
	freshHash func() hash.Hash
	eds       *rsmt2d.ExtendedDataSquare
	opts      *nmt.Options
}

// Commitment returns the root of a selected row or column using a configured
// name space merkle tree. It also fullfills it's portion of the
// rsmt2d.VectorCommiterProver interface.
func (v VCP) Commitment(a rsmt2d.Axis, idx uint) []byte {
	// push all the data onto an nmt
	tree := New(v.freshHash(), v.setters...)
	// use data from the selected row or column of the extended data square 
	leaves := v.fetchLeaves(a, idx)
	// push data to the tree
	for _, leaf := range leaves {
		err := tree.Push(namespace.NewPrefixedData(tree.NamespaceSize(), leaf))
		if err != nil {
			panic(
				fmt.Sprintf("invalid data; could not push share to tree: %#v, err: %v",
					leaf,
					err,
				),
			)
		}
	}
	return tree.Root().Bytes()
}
```
It should be noted that this experimental version requires that the extended data square be prefixed with `namespace.ID`s. [(which was done here)](https://github.com/evan-forbes/rsmt2d/blob/02d7966e58d9e20d7f49c8b287e46c99451e7abd/extendeddatasquare.go#L43)

At this point, we can define some quality of life functions that now work with all VCP implementations
```go
// Commitments uses the provided VectorCommitmentProver to collect commitments
// from a selected axis
func Commitments(a Axis, vcp VectorCommitmentProver, width uint) [][]byte {
	commits := make([][]byte, width)
	for i := uint(0); i < width; i++ {
		commits[i] = vcp.Commitment(a, i)
	}
	return commits
}
```
and finally we can get a decent idea of how these changes would look like in lazyledger-core
```go
// fillDataAvailabilityHeader fills in any remaining DataAvailabilityHeader fields
// that are a function of the block data.
func (b *Block) fillDataAvailabilityHeader() {
	namespacedShares := b.Data.computeShares()
	if len(namespacedShares) == 0 {
		// no shares -> no row/colum roots -> hash(empty)
		b.DataHash = b.DataAvailabilityHeader.Hash()
		return
	}
	rawData, names := namespacedShares.RawShares(), namespacedShares.IDs()
	// TODO(ismail): for better efficiency and a
	// larger number shares we should switch to the rsmt2d.LeopardFF16 codec:
	// erasure the data and assign appropriate namespace IDs after
	extendedDataSquare, err := rsmt2d.ComputeNamedExtendedDataSquare(rawData, names, rsmt2d.RSGF8)
	if err != nil {
		panic(fmt.Sprintf("unexpected error: %v", err))
	}
	// create an nmt VectorCommitmentProver and use the size of the first namespaced share
	vcp := nmt.NewVCP(extendedDataSquare, sha256.New, nmt.NamespaceIDSize(int(namespacedShares[0].ID.Size())))

	// build the data availability header
	width := extendedDataSquare.Width()
	colRoots := rsmt2d.Commitments(rsmt2d.Column, vcp, width)
	rowRoots := rsmt2d.Commitments(rsmt2d.Row, vcp, width)

	// set the dah roots
	b.DataAvailabilityHeader.ColumnRoots = NmtRootsFromBytes(colRoots)
	b.DataAvailabilityHeader.RowsRoots = NmtRootsFromBytes(rowRoots)
	//  create and set the root hash
	b.DataHash = merkle.HashFromByteSlices(append(rowRoots, colRoots...))
}
```
Overall, these experimental changes are purely organizational, and only achieve opinionated aesthetic changes... so they're not really needed. Especially considering that, as far as I know, there are no plans to have multiple implementations of name spaced merkle trees. There's also a small amount of overhead with O(n) complexity introduced due to the conversion of `namespace.ID`s to `[]byte` and then back to `namespace.ID` when passing data through the nmt VectorCommitmentProver implementation. If there's continued interest, I'd be more than happy to keep investigating, and see what can be done to more closely weave the nmt and rsmt2d packages into lazyledger-core.

Commits:
New interface:
rsmt2d: https://github.com/evan-forbes/rsmt2d/commit/02d7966e58d9e20d7f49c8b287e46c99451e7abd
Experimental nmt VCP:
nmt: https://github.com/evan-forbes/nmt/commit/1ca1d871fb9e7d729243242f3f701d3b6b7bd14e
lazyledger-core: https://github.com/evan-forbes/lazyledger-core/commit/03398e3520d765da8f49324b9436474dad7e0552

musalbas:
Thanks for this @evan-forbes. Requiring a wrapper around EDS seems like an overly complex pattern. Why not simply create an interface for a merkle tree, and allow the EDS to be initialised with any merkle tree object that implements that interface?

evan-forbes:
The VCP interfaces allows for more freedom in aggregating data for a proof or commitment, specifically for cached trees. The `ExtendedDataSquare` would have to assume that each implementation is not cached, so it would create a new tree and push data each time a proof needs to be generated, where VCP does not make that assumption.

musalbas:
I think the caching mechanism should probably be implemented on the EDS, rather in the tree library itself.

datasquare.go
```diff
@@ -190,34 +177,26 @@ func (ds *dataSquare) ColumnRoots() [][]byte {
...
 
 	for i := uint(0); i < ds.width; i++ {
 		tree.Push(data[i])
 	}
-
-	merkleRoot, proof, proofIndex, numLeaves := tree.Prove()
+	// TODO(ismail): check for overflow when casting from uint -> int
+	merkleRoot, proof, proofIndex, numLeaves := tree.Prove(int(x))
 	return merkleRoot, proof, uint(proofIndex), uint(numLeaves), nil
 }
```
evan-forbes, line 202: I think the `dataSquare` type is going to need to expose a (some) method(s) for creating a Proof

evan-forbes:
I can see that working well. It might be a tad confusing to the user, as any caching that requires access to the tree's internals, such as the caching that occurs in the current version of nmt, would obviously be separate from the type that would be occurring in the EDS. Then again, it might be desirable to keep both types of caching separate. 

There's also still the option to keep what is currently written, and not include an interface for the time being. We might be risking a premature abstraction, as the entire use of rsmt2d and nmt in lazyledger-core is currently confined to the `fillDataAvailabilityHeader` method of `block`, and there is only a single nmt implementation to use.

liamsi:
Very good discussion! I like the idea of generalizing this to Evan's proposed VCP interface but it's another very good point that @evan-forbes  brought up himself: this might indeed be a premature absraction. For now, I just removed the other single-function interface the `TreeCreator` tentatively to be renamed sth-sth-factory (and kept the tree interface). 
It's now a function that returns a `Tree`. While it does the same thing, I did not really feel comfortable introducing another interface.

tree.go
liamsi: Got rid of it in https://github.com/lazyledger/rsmt2d/pull/9/commits/51a97d9ef976d3c58b27de1c26678f1f4ac28a73

tree.go
```diff
@@ -7,30 +7,25 @@ import (
...
 	Prove(idx int) (merkleRoot []byte, proofSet [][]byte, proofIndex uint64, numLeaves uint64)
 	Root() []byte
 }
 
-type TreeCreator interface {
-	New() Tree
-}
-
-var _ TreeCreator = &defaultTreeCreator{}
 var _ Tree = &DefaultTree{}
```
musalbas, line 17: How would developers use their own tree implementation for the library? Create a struct that implements Tree, then supply a function that creates a new tree?

tree.go
```diff
@@ -7,30 +7,25 @@ import (
...
 
-type defaultTreeCreator struct{}
+// TreeFn creates a fresh Tree instance to be used as the Merkle inside of rsmt2d.
+type TreeFn = func() Tree
```
musalbas, line 20: Wonder if TreeInit or TreeConstructor would be a more descriptive name than TreeFn?

musalbas (approved):

tree.go
liamsi: > Create a struct that implements Tree, then supply a function that creates a new tree?

Yes, exactly.

tree.go
liamsi: Renamed to `TreeConstructorFn` in 589b0b3.

PR #18, opened 2021-03-23
implement lazy root computation

### LINKED ISSUE #17: Lazily compute row and column roots when repairing the square
When repairing the square, we are currently generating roots of each column and row even if there is nil data in that row or col. This is a problem because because if the underlying tree doesn't know how to handle `nil` values, then it has no choice but to panic. While there are quite a few ways to fix this, I think the fastest way would be to only compute the root of columns and rows that have no nil values. In the context of `RepairDatasquare`, this means having the ability to generate roots of single columns and rows.

This must be fixed before [#235](https://github.com/lazyledger/lazyledger-core/pull/235) and [#232](https://github.com/lazyledger/lazyledger-core/pull/232) can be completed.

evan-forbes:
this PR is a slight refactor, that adds the ability to compute the root of a single row or column. This functionality is useful repairing the square without feeding the underlying `rsmt2d.Tree` nil values when not needed.

While the current root caching mechanism is used, no attempt was made to cache individual row or column roots. This means that if any element of the square changes, then like the current implementation, each root must be recomputed. :man_shrugging:  

I'm keeping this as a draft for now, as more changes could be needed to finish [#235](https://github.com/lazyledger/lazyledger-core/pull/235)

closes #17

### DIFF
--- a/datasquare_test.go
+++ b/datasquare_test.go
@@ -76,7 +76,7 @@ func TestLazyRootGeneration(t *testing.T) {
 	var rowRoots [][]byte
 	var colRoots [][]byte
 
-	for i := uint(0); i < 2; i++ {
+	for i := uint(0); i < square.width; i++ {
 		rowRoots = append(rowRoots, square.RowRoot(i))
 		colRoots = append(rowRoots, square.ColRoot(i))
 	}
--- a/datasquare_test.go
+++ b/datasquare_test.go
@@ -67,6 +67,28 @@ func TestRoots(t *testing.T) {
 	}
 }
 
+func TestLazyRootGeneration(t *testing.T) {
+	square, err := newDataSquare([][]byte{{1}, {2}, {3}, {4}}, NewDefaultTree)
+	if err != nil {
+		panic(err)
+	}
+
+	var rowRoots [][]byte
+	var colRoots [][]byte
+
+	for i := uint(0); i < 2; i++ {
+		rowRoots = append(rowRoots, square.RowRoot(i))
+		colRoots = append(rowRoots, square.ColRoot(i))
+	}
+
+	square.computeRoots()
+
+	if !reflect.DeepEqual(square.rowRoots, rowRoots) && !reflect.DeepEqual(square.columnRoots, colRoots) {
+		t.Error("RowRoot or ColumnRoot did not produce identical roots to computeRoots")
+	}
+
+}
+
 func TestProofs(t *testing.T) {
 	result, err := newDataSquare([][]byte{{1, 2}, {3, 4}, {5, 6}, {7, 8}}, NewDefaultTree)
 	if err != nil {
--- a/datasquare_test.go
+++ b/datasquare_test.go
@@ -86,7 +86,30 @@ func TestLazyRootGeneration(t *testing.T) {
 	if !reflect.DeepEqual(square.rowRoots, rowRoots) && !reflect.DeepEqual(square.columnRoots, colRoots) {
 		t.Error("RowRoot or ColumnRoot did not produce identical roots to computeRoots")
 	}
+}
 
+func TestRootAPI(t *testing.T) {
+	square, err := newDataSquare([][]byte{{1}, {2}, {3}, {4}}, NewDefaultTree)
+	if err != nil {
+		panic(err)
+	}
+
+	for i := uint(0); i < square.width; i++ {
+		if !reflect.DeepEqual(square.RowRoots()[i], square.RowRoot(i)) {
+			t.Errorf(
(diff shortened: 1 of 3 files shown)


datasquare_test.go
```diff
@@ -67,6 +67,28 @@ func TestRoots(t *testing.T) {
...
+func TestLazyRootGeneration(t *testing.T) {
+	square, err := newDataSquare([][]byte{{1}, {2}, {3}, {4}}, NewDefaultTree)
+	if err != nil {
+		panic(err)
+	}
+
+	var rowRoots [][]byte
+	var colRoots [][]byte
+
+	for i := uint(0); i < 2; i++ {
```
liamsi, line 79: make `2` a var that has a meaningful name.

liamsi (commented):
So the only difference is that we extend the API to be able to compute specific row/col roots on demand. Looks like a good change anyway.

datasquare_test.go
```diff
@@ -67,6 +67,28 @@ func TestRoots(t *testing.T) {
...
+		rowRoots = append(rowRoots, square.RowRoot(i))
+		colRoots = append(rowRoots, square.ColRoot(i))
+	}
+
+	square.computeRoots()
```
liamsi, line 84: It would be good to have a test that only uses public API, too.

e.g.: assert that calling `RowRoots()[i]` yields the same result as calling `RowRoot(i)` etc

extendeddatacrossword.go
```diff
@@ -229,11 +228,26 @@ func (eds *ExtendedDataSquare) prerepairSanityCheck(rowRoots [][]byte, columnRoo
 	for i := uint(0); i < eds.width; i++ {
 		rowMask := mask.RowView(int(i))
 		columnMask := mask.ColView(int(i))
-		if (vecIsTrue(rowMask) && !bytes.Equal(rowRoots[i], eds.RowRoots()[i])) || (vecIsTrue(columnMask) && !bytes.Equal(columnRoots[i], eds.ColumnRoots()[i])) {
-			return errors.New("bad roots input")
+		rowMaskIsVec := vecIsTrue(rowMask)
+		columnMaskIsVec := vecIsTrue(columnMask)
+
+		// if there's no missing data in the this row
+		if !checkForNil(eds.Row(i)) {
```
liamsi, line 235: What do you think of renaming the function to `noMissingData` and removing the `!`?

datasquare_test.go
evan-forbes: good call. resolved https://github.com/lazyledger/rsmt2d/pull/18/commits/06294d583ac8f33a01085da3f578ba40f4151c2a

extendeddatacrossword.go
evan-forbes: that is definitely easier to read, thanks! resolved https://github.com/lazyledger/rsmt2d/pull/18/commits/193de535108c7547b5263fecec7716545bdde58f

datasquare_test.go
evan-forbes: forgot which PR this snuck into but fixed https://github.com/lazyledger/rsmt2d/blob/06294d583ac8f33a01085da3f578ba40f4151c2a/datasquare_test.go#L79

extendeddatacrossword.go
```diff
@@ -184,12 +183,12 @@ func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][
...
 								if mode == row {
 									adjMask := mask.ColView(int(j))
-									if vecNumTrue(adjMask) == adjMask.Len()-1 && !bytes.Equal(eds.ColumnRoots()[j], columnRoots[j]) {
+									if vecNumTrue(adjMask) == adjMask.Len()-1 && !bytes.Equal(eds.ColRoot(j), columnRoots[j]) {
 										return &ByzantineColumnError{j, edsBackup}
 									}
 								} else if mode == column {
 									adjMask := mask.RowView(int(j))
-									if vecNumTrue(adjMask) == adjMask.Len()-1 && !bytes.Equal(eds.RowRoots()[j], rowRoots[j]) {
+									if vecNumTrue(adjMask) == adjMask.Len()-1 && !bytes.Equal(eds.RowRoot(j), rowRoots[j]) {
```
liamsi, line 191: Orthogonal to this PR but we should really refactor this. The massive indentation / nesting alone cries for splitting this up into multiple methods...
```suggestion
									if vecNumTrue(adjMask) == adjMask.Len()-1 && !bytes.Equal(eds.RowRoot(j), rowRoots[j]) {
```

liamsi (approved):
🟢 👍🏼  🤸🏼‍♂️ 🙌🏼

PR #20, opened 2021-03-26
add cell index to the Push method of Tree

### LINKED ISSUE #19: Upated the Tree interface to communicate row and column position
Per [comment](https://github.com/lazyledger/lazyledger-core/issues/230#issuecomment-808249600), we're going to try updating the `Tree` interface so that row and column position are passed when calling the `Push` method.

evan-forbes:
This PR modifies the `Tree` interface to pass along the information needed to identify which cell is being pushed.

closes #19

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -40,7 +40,7 @@ func NewDefaultTree() Tree {
 	}
 }
 
-func (d *DefaultTree) Push(data []byte, idx CellIndex) {
+func (d *DefaultTree) Push(data []byte, _idx CellIndex) {
 	// ignore the idx, as this implementation doesn't need that info
 	d.leaves = append(d.leaves, data)
 }
--- a/tree.go
+++ b/tree.go
@@ -10,19 +10,26 @@ import (
 // TreeConstructorFn creates a fresh Tree instance to be used as the Merkle inside of rsmt2d.
 type TreeConstructorFn = func() Tree
 
+// Axis indicates whether a Column or Row is being used
 type Axis bool
 
 const (
+	// Row indicates the use of the x axis of a data square
 	Row Axis = true
+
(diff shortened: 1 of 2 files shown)


tree.go
```diff
@@ -32,7 +40,8 @@ func NewDefaultTree() Tree {
 	}
 }
 
-func (d *DefaultTree) Push(data []byte) {
+func (d *DefaultTree) Push(data []byte, idx CellIndex) {
```
liamsi, line 43: It's common to prefix unused vars with `_` or you can simply use `_`.
```suggestion
func (d *DefaultTree) Push(data []byte, _idx CellIndex) {
```
or 
```go
func (d *DefaultTree) Push(data []byte, _ CellIndex) {
```

tree.go
```diff
@@ -10,8 +10,16 @@ import (
 // TreeConstructorFn creates a fresh Tree instance to be used as the Merkle inside of rsmt2d.
 type TreeConstructorFn = func() Tree
 
+// CellIndex contains all information needed to identify the cell that is being
+// pushed
+type CellIndex struct {
+	AxisIndex uint
+	CellIndex uint
```
liamsi, line 17: Is it on purpose that the type CellIndex contains a field with the same name as the type (CellIndex)?

evan-forbes: whoops, changed the type `CellIndex` -> `SquareIndex`, and dropped the `Index` of the fields. Happy to change it something else if anyone has some ideas https://github.com/lazyledger/rsmt2d/pull/20/commits/1f35c00fc7dbd9b0e5dbecb4001d9acefdf9f18c

liamsi (approved):
👍🏼

PR #24, opened 2021-04-02
code hygiene: idiomatic error naming

liamsi:
I've spend some time looking into refactoring and cleaning up rsmt2d. This is the first of a few smaller PRs in this direction:
- prefix errors with `Err`
- remove tracking EDS in errors

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -12,23 +12,21 @@ const (
 	column = 1
 )
 
-// ByzantineRowError is thrown when there is a repaired row does not match the expected row merkle root.
-type ByzantineRowError struct {
-	RowNumber      uint
-	LastGoodSquare ExtendedDataSquare
+// ErrByzantineRow is thrown when there is a repaired row does not match the expected row merkle root.
+type ErrByzantineRow struct {
+	RowNumber uint
 }
 
-func (e *ByzantineRowError) Error() string {
+func (e *ErrByzantineRow) Error() string {
 	return fmt.Sprintf("byzantine row: %d", e.RowNumber)
 }
 
-// ByzantineColumnError is thrown when there is a repaired column does not match the expected column merkle root.
-type ByzantineColumnError struct {
-	ColumnNumber   uint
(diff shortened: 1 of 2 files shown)


adlerjohn (approved):

evan-forbes (approved):
👍

extendeddatacrossword.go
```diff
@@ -12,34 +13,27 @@ const (
...
 
-func (e *ByzantineRowError) Error() string {
+func (e *ErrByzantineRow) Error() string {
 	return fmt.Sprintf("byzantine row: %d", e.RowNumber)
 }
 
-// ByzantineColumnError is thrown when there is a repaired column does not match the expected column merkle root.
-type ByzantineColumnError struct {
-	ColumnNumber   uint
-	LastGoodSquare ExtendedDataSquare
```
liamsi, line 28: Is it really OK to remove this `LastGoodSquare`? It's currently not really used but I'm not sure if this will be needed to proof that this column was actually byzantine? (cc @adlerjohn)

adlerjohn: I don't really see the need for returning a backup copy?

liamsi: Yeah, that's what I thought too. Thanks for confirming. The data is in the hands of the caller anyways. If required, the caller can construct everything from the row/col index and the data again (not sure if they need to in any case though).

It would help if the spec actually explained under what circumstances nodes and (which nodes?) would construct a BadEncodingProof from what data. Should I open an issue about that?

adlerjohn: > Should I open an issue about that?

Maybe a comment in https://github.com/lazyledger/lazyledger-specs/issues/23 ? I'm going to start working on Evidence relatively soon.

liamsi: https://github.com/lazyledger/lazyledger-specs/issues/23#issuecomment-812879786
Is that clear enough?

adlerjohn: Yes, thanks.

PR #26, opened 2021-04-04
Add Apache-2.0 license

adlerjohn:
cc @musalbas


liamsi (approved):
Thanks, I think it makes sense to use apache2 accross all repositories.

Issue #22, opened 2021-04-02
Proposal: use light-weight bitmask instead of gonums's mat.Dense


liamsi:
This float64 slice
https://github.com/lazyledger/rsmt2d/blob/ef1d6c54461edaa428b5720835b6b522c18defff/extendeddatacrossword.go#L52-L63
is used to initialize a dense matrix: https://github.com/lazyledger/rsmt2d/blob/ef1d6c54461edaa428b5720835b6b522c18defff/extendeddatacrossword.go#L82

It looks like this could be vastly improved by using some form of 2d-bitmask to look up if a cell in a row or column is present or not, or if a row / column is incomplete. So instead of 64 bits per element, we could reduce that to 1 bit per element.

We can either look for a small library out there to use or we can write this our selves.

liamsi:
This is how this could look like: https://github.com/lazyledger/rsmt2d/blob/9454d9b9c7ed4d9cf5fcbb3ffe3121ab304d9585/bit_matrix.go#L17-L30
used here: https://github.com/lazyledger/rsmt2d/blob/9454d9b9c7ed4d9cf5fcbb3ffe3121ab304d9585/extendeddatacrossword.go#L50

PR #27, opened 2021-04-05
Update and tidy `go.mod`

adlerjohn:
- Tidy `go.mod`
- ~~Remove unused dependencies~~
- Update dependencies
- Update Go version to 1.15


go.mod
```diff
@@ -1,17 +1,21 @@
 module github.com/lazyledger/rsmt2d
 
-go 1.14
+go 1.15
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/kr/text v0.2.0 // indirect
```
liamsi, line 7: Can we at least understand what caused this to be imported? If it is gonum, we will remove gonum with #22.

go.mod
```diff
@@ -1,17 +1,21 @@
...
 	// only needed if built with go build -tags leopard
-	github.com/lazyledger/go-leopard v0.0.0-20200604113236-298f93361181
+	github.com/lazyledger/go-leopard v0.0.0-20200724211609-50ec4b3fab41
```
liamsi, line 9: We should also tag go-leopard. I'll revisit that repository soon.

go.mod
```diff
@@ -1,17 +1,21 @@
...
-	github.com/spacemonkeygo/errors v0.0.0-20171212215202-9064522e9fd1 // indirect
-	github.com/stretchr/testify v1.5.1
-	github.com/vivint/infectious v0.0.0-20190108171102-2455b059135b
-	golang.org/x/net v0.0.0-20200520182314-0ba52f642ac2 // indirect
-	golang.org/x/sys v0.0.0-20200519105757-fe76b779f299 // indirect
-	gonum.org/v1/gonum v0.7.0
-	gopkg.in/yaml.v2 v2.3.0 // indirect
+	github.com/stretchr/testify v1.7.0
+	github.com/vivint/infectious v0.0.0-20200605153912-25a574ae18a3
+	gitlab.com/NebulousLabs/errors v0.0.0-20200929122200-06c536cf6975 // indirect
```
liamsi, line 13: It's really surprising to see these updated while the merkle tree stayed the same 🤔

liamsi (approved):
👍🏼

go.mod
adlerjohn: For posterity: this is dependency of `yaml` v3, which is a dependency of `testify`

PR #30, opened 2021-04-05
Migrate to GitHub Actions

adlerjohn:
Fixes #29

PR #31, opened 2021-04-05
Update readme badges

adlerjohn:
Update badges for Actions from #30.

Rendered: https://github.com/adlerjohn/rsmt2d/blob/adlerjohn/update_readme/README.md


liamsi (approved):
👍🏼

PR #32, opened 2021-04-05
Fix enums to use iota

adlerjohn:
https://www.iota.org/

### DIFF
--- a/codecs.go
+++ b/codecs.go
@@ -12,8 +12,8 @@ type CodecType int
 const (
 	// RSGF8 represents Reed-Solomon codecType with an 8-bit Finite Galois Field (2^8)
 	RSGF8       CodecType = iota
-	LeopardFF8  CodecType = 1
-	LeopardFF16 CodecType = 2
+	LeopardFF8  CodecType = iota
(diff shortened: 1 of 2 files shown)


liamsi:
The notification with the URL scare me for a second 🤣

codecs.go
```diff
@@ -12,8 +12,8 @@ type CodecType int
 const (
 	// RSGF8 represents Reed-Solomon codecType with an 8-bit Finite Galois Field (2^8)
 	RSGF8       CodecType = iota
-	LeopardFF8  CodecType = 1
-	LeopardFF16 CodecType = 2
+	LeopardFF8  CodecType = iota
+	LeopardFF16 CodecType = iota
```
liamsi, line 16: Can this be simplified s.t. only the first line says iota?
Same below.

adlerjohn: These cannot, since `CodecType` is a custom type. The compiler needs an actual value assigned. The other ones are just `int`s, so they can.

adlerjohn: Fixed in 888fd64.

liamsi (approved):
👍🏼

Issue #12, opened 2021-01-11
Clarify public API


liamsi:
The current public facing API:
![image](https://user-images.githubusercontent.com/7441766/104191904-69759e00-541e-11eb-880f-560c655427d8.png)

Should we:
 - add methods to the extended data square? For instance: serialization and deserialzation (e.g. from and to protobuf), (square/orig. data) Width
 - clarify the names of the existing Encode/Decode methods (e.g. by exposing the methods of the `Codec` interface)
 - decouple the erasure coding step from the row/col roots step

related: #11

liamsi:
For LazyLegder, we need to modify the erasure coded shares before they are committed to as row / col roots. Namely, we need to prefix the orig. data and the parity shares with their namespaces respectively. This is already possible using the public API by computing the row / column roots externally. But if[ we want to use the rsmt2d](https://github.com/lazyledger/lazyledger-core/blob/78d11a92cb10f91b878d5d26558f33c318bd4508/types/block.go#L220-L222) for this instead, we need to make it possible to transform the shares before they are committed to internally.

See for example: https://github.com/lazyledger/lazyledger-core/blob/4256e2090ba71873ee4698353de7d6646db97907/p2p/ipld/plugin/plugin_test.go#L116-L129

adlerjohn:
Additional: https://github.com/lazyledger/rsmt2d/pull/56#discussion_r622810103

liamsi:
Another (potentially last?) point to address with the is how the library handles proofs:

Currently, we extracted a tree interface with a `Prove` method:
https://github.com/lazyledger/rsmt2d/blob/5a692bff96defabeaf8c928510c1894255e97761/tree.go#L22-L23

liamsi:
Closinf in favour of:
https://github.com/celestiaorg/rsmt2d/issues/82
https://github.com/celestiaorg/rsmt2d/issues/83
https://github.com/celestiaorg/rsmt2d/issues/84
https://github.com/celestiaorg/rsmt2d/issues/85

And other improvements that were already done:

https://github.com/celestiaorg/rsmt2d/issues/78
https://github.com/celestiaorg/rsmt2d/issues/77

PR #33, opened 2021-04-05
Introduce bitMatrix to replace mat.Dense

### LINKED ISSUE #22: Proposal: use light-weight bitmask instead of gonums's mat.Dense

liamsi:
This PR introduces a structure that replaces mat.Dense and gets rid of the gonum dependency.
~~Hence, it is half the story of~~ closes #22.

It could use some light fuzzing but given that this is essentially a simple bit-mask I don't think that is urgent.

~~Can open a separate PR to actually use the structure, or, replace mat.Dense in here.~~


bit_matrix.go
```diff
@@ -0,0 +1,100 @@
+package rsmt2d
+
+import "fmt"
+
+type bitMatrix struct {
```
liamsi, line 5: Maybe, I should note that the thing is not go-rountine-safe. That doesn't matter at all right now but I know that @evan-forbes experimented with paralleization in https://github.com/lazyledger/rsmt2d/issues/5 already.

liamsi:
As requested, this PR also replaces gonum now. It would have been nice to measure the impact with benchmarks but we already know the ballpark of improvment (see #22).

bit_matrix.go
```diff
@@ -0,0 +1,98 @@
...
+	bits := squareSize * squareSize
+	return bitMatrix{mask: make([]uint64, (bits+63)/64), squareSize: squareSize}
+}
+
+// i = rowIndex*squareSize+colIdx
+func (bm bitMatrix) SetFlat(i int) {
+	bm.mask[i/64] |= uint64(1) << uint(i%64)
+}
+
+func (bm bitMatrix) Get(row, col int) bool {
```
evan-forbes, line 20: Do these methods have to be exported?

liamsi: Not really, no. Everything is in a single flat root package here anyways and even the type itself is not exported `bitMatrix`. Both the type and these methods would need to be exported if this was moved into a separaste go package. 
I only exported all these methods so it is clear which are meant to be used as an API, in other words, to communicate the surface of this if it was a library. The caller can currently bypass that anyways: they could directly mingle with the underlying `mask` field.

evan-forbes: I can definitely dig that, nice

bit_matrix.go
adlerjohn: What's the alternative here? How much work would it be to make this thread-safe? Is there not a standard library for this that is thread-safe?

liamsi: > How much work would it be to make this thread-safe?

Basically adding mutexes for accessing the underlying field. 

> Is there not a standard library for this that is thread-safe?

Not to my knowledge.

Also, we currently do not need this to be thread-safe as we aren't accessing any of this via go-routines.

bit_matrix.go
```diff
@@ -0,0 +1,98 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func (bm bitMatrix) NumOnesInRow(r int) int {
+	var counter int
+	for i := 0; i < bm.squareSize; i++ {
+		if bm.Get(r, i) {
```
adlerjohn, line 53:
```suggestion
	for c := 0; c < bm.squareSize; c++ {
		if bm.Get(r, c) {
```

bit_matrix.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+
+	return counter
+}
+
+func (bm bitMatrix) NumOnesInCol(c int) int {
+	var counter int
+	for i := 0; i < bm.squareSize; i++ {
+		if bm.Get(i, c) {
```
adlerjohn, line 64:
```suggestion
	for r := 0; r < bm.squareSize; r++ {
		if bm.Get(r, c) {
```

bit_matrix.go
```diff
@@ -0,0 +1,98 @@
...
+
+func newBitMatrix(squareSize int) bitMatrix {
+	bits := squareSize * squareSize
+	return bitMatrix{mask: make([]uint64, (bits+63)/64), squareSize: squareSize}
+}
+
+// i = rowIndex*squareSize+colIdx
+func (bm bitMatrix) SetFlat(i int) {
+	bm.mask[i/64] |= uint64(1) << uint(i%64)
+}
```
adlerjohn, line 18:
```suggestion
// idx = rowIndex*squareSize+colIdx
func (bm bitMatrix) SetFlat(idx int) {
	bm.mask[idx/64] |= uint64(1) << uint(idx%64)
}
```

adlerjohn (commented):
Is there a reason not to use [`big.Int`](https://golang.org/pkg/math/big/#Int) as the backing type instead of a `[]uint64`?

evan-forbes:
I ran this quick fuzzing test locally in lazyledger-core, which continually runs the [`TestBlockRecovery`](https://github.com/lazyledger/lazyledger-core/blob/93c88a1ac35ce6d0e66554f635424e808c7efb9b/p2p/ipld/read_test.go#L159) test. With the master version of rsmt2d, this test passes, but with this branch of rsmt2d, it fails.
```go
func TestFuzzTestBlockRecovery(t *testing.T) {
	ctx, _ := context.WithTimeout(context.Background(), time.Minute)
	for {
		select {
		case <-ctx.Done():
			return
		default:
			TestBlockRecovery(t)
		}
	}
}
```

```
    --- FAIL: TestFuzzTestBlockRecovery/missing_1/2_shares#295 (0.00s)
        read_test.go:220: 
                Error Trace:    read_test.go:220
                Error:          Received unexpected error:
                                failed to solve data square
                Test:           TestFuzzTestBlockRecovery/missing_1/2_shares#295
```

I'm not sure if it's just `TestBlockRecover` that is wrong, or if it's something here.

extendeddatacrossword.go
```diff
@@ -73,172 +71,183 @@ func RepairExtendedDataSquare(
...
 				}
 			}
 		}
 
-		if solved {
-			break
-		} else if !progressMade {
+		if !progressMade {
 			return ErrUnrepairableDataSquare
 		}
```
evan-forbes, line 186: Ahh I think this is issue for the failing lazyledger-core test mentioned in the other comment, the square is being repaired fine, but an error is being returned.

liamsi: Good catch!! I'll revert these changes in the for-loop here. They are orthogonal to the bit-mask anyways.

liamsi: Done in https://github.com/lazyledger/rsmt2d/pull/33/commits/3eca5806440382bbe899fe1d66700fedc9c6c2aa.
I'll retest with the above test to make sure that this fixed the issue.

bit_matrix.go
liamsi: addressed all these renamings in https://github.com/lazyledger/rsmt2d/pull/33/commits/3eca5806440382bbe899fe1d66700fedc9c6c2aa too

liamsi:
> Is there a reason not to use big.Int as the backing type instead of a []uint64?

Is there a good reason to do so? I think the main reason is that you can not directly operate on the underlying []uint and would need to use [`Bits()`](https://golang.org/pkg/math/big/#Int.Bits) which returns almost the same as we are already using. Then, big.Int is signed which do not need. Also, the underlying "words" used are `Word uint` and `uint`'s size depends on the platform used: https://golang.org/pkg/builtin/#uint
> uint is an unsigned integer type that is _at least_ 32 bits in size.

extendeddatacrossword.go
liamsi: An alternative could be to change that if-else-if to: `if !progressMade && !solved` (or `!(progressMade || solved)`) and make the for loop `for !solved`. That should be equivalent. I think the for-loop would be easier to grasp directly (repeat until solved).

liamsi: Either way, thanks for finding that issue and for pointing me to the solution even :-)

datasquare.go
```diff
@@ -16,7 +16,7 @@ type dataSquare struct {
 
 func newDataSquare(data [][]byte, treeCreator TreeConstructorFn) (*dataSquare, error) {
 	width := int(math.Ceil(math.Sqrt(float64(len(data)))))
-	if int(math.Pow(float64(width), 2)) != len(data) {
+	if width*width != len(data) {
```
liamsi, line 19: I've sneaked in another unrelated change here: `int(math.Pow(float64(width), 2))` is just a complicated way of computing `width*width`.

bit_matrix.go
liamsi: IMO, we should put stuff like this into an internal package: https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface
But maybe we can tackle this together with #12 later.

datasquare.go
adlerjohn: Saw that, and approve.

adlerjohn (approved):

liamsi:
can someone with write access approve the PR (or request changes): cc @musalbas @evan-forbes

evan-forbes (approved):
awesome :+1: :+1:

Issue #25, opened 2021-04-03
add back LastGoodSquare to errors


liamsi:
Revert removing the LastGoodSquare from the byzantine row/col errors (in #24).

copy-pasting @musalbas' clarification here:

> It's so that fraud proofs can be generated for invalid codes, using the eds as much as it can be reconstructed.
> It's not OK to remove LastGoodSquare

@adlerjohn  had some ideas on how to add it without doing a deep-copy:
> Basically, instead of corrupting the EDS here https://github.com/lazyledger/rsmt2d/blob/6610ba486a29d90dd0faa51f539da1415777a8f2/extendeddatacrossword.go?ts=4#L152-L158 then doing the check https://github.com/lazyledger/rsmt2d/blob/6610ba486a29d90dd0faa51f539da1415777a8f2/extendeddatacrossword.go?ts=4#L161-L170, do the check first then set the EDS in a way that's guaranteed correct. The check would have to compute a root based on existing leaves and a new set of repaired leaves, but that's just copy-pasting this logic https://github.com/lazyledger/rsmt2d/blob/6610ba486a29d90dd0faa51f539da1415777a8f2/extendeddatacrossword.go?ts=4#L152-L158.

But we can also just add it back as before and do the optimization later.

More important than above mentioned optimzations are test-cases (additional to go-doc comments) that document the purpose of the `LastGoodSquare` fields, for example by generating above mentioned fraud proofs (also related to #12).

PR #35, opened 2021-04-06
Add benchmarks for repairing

### LINKED ISSUE #23: Add benchmarks for RepairExtendedDataSquare too
The only benchmarks we have in this repository are those added by @evan-forbes in #14.

It would be good if we also benchmarked repairing the square (allocs and ops). This would help deciding on some changes like #22 and others. We should first aim for better readability though.

adlerjohn:
Fixes #23


codec_test.go
```diff
@@ -22,11 +22,12 @@ func TestCodec_String(t *testing.T) {
 }
 
 func BenchmarkEncoding(b *testing.B) {
+	fmt.Println(benchmarkDivider)
```
liamsi, line 25: IMO, this divider is not necessary. Also, I'm not sure if tools like benchstat can deal with that.

codec_test.go
```diff
@@ -22,11 +22,12 @@ func TestCodec_String(t *testing.T) {
...
 	// generate some fake data
 	data := generateRandData(128)
 	for _codecType := range codecs {
 		b.Run(
-			fmt.Sprintf("Encoding 128 shares using %s", _codecType),
+			fmt.Sprintf("Encoding 128 shares using %s\n", _codecType),
```
liamsi, line 30: Optionally revert that change:
```suggestion
			fmt.Sprintf("Encoding 128 shares using %s", _codecType),
```

The newline doesn't make much sense here as it would just result in an `_` (at least I think this is how go treats whitespaces here).

liamsi: Yes, can confirm. Without the newline:
```
BenchmarkEncoding
BenchmarkEncoding/Encoding_128_shares_using_RSGF8
BenchmarkEncoding/Encoding_128_shares_using_RSGF8-12         	    4750	    237889 ns/op
```
with newline:
```
BenchmarkEncoding
BenchmarkEncoding/Encoding_128_shares_using_RSGF8_
BenchmarkEncoding/Encoding_128_shares_using_RSGF8_-12         	    4092	    250458 ns/op
```

codec_test.go
adlerjohn: The divider is there for the other benchmarks, so I just made everything even. Do we want to remove all the dividers? I'm impartial.

codec_test.go
adlerjohn: Ah that was me playing around with the strings and forgetting to remove the `\n`. Thanks.

codec_test.go
liamsi: Yeah, let's remove all of them.

adlerjohn: Removed in 3910e3b.

extendeddatacrossword_test.go
```diff
@@ -110,3 +111,44 @@ func TestRepairExtendedDataSquare(t *testing.T) {
...
 	}
 }
+
+func BenchmarkRepair(b *testing.B) {
+	for _codecType := range codecs {
+		bufferSize := 64
+		ones := bytes.Repeat([]byte{1}, bufferSize)
+		twos := bytes.Repeat([]byte{2}, bufferSize)
+		threes := bytes.Repeat([]byte{3}, bufferSize)
+		fours := bytes.Repeat([]byte{4}, bufferSize)
```
liamsi, line 121: why don't we use random data? Is 64 the bufferSize that we want to optimize for? I thought shares will 256 bytes rather. Also the square size should be close to our estimated relevant square sizes. Maybe 3 or 4 different square sizes would make the most sense (almost empty, small, medium, really large).

extendeddatacrossword_test.go
```diff
@@ -110,3 +111,44 @@ func TestRepairExtendedDataSquare(t *testing.T) {
...
+		}, _codecType, NewDefaultTree)
+		if err != nil {
+			panic(err)
+		}
+
+		flattened := original.flattened()
+		flattened[0], flattened[2], flattened[3] = nil, nil, nil
+		flattened[4], flattened[5], flattened[6], flattened[7] = nil, nil, nil, nil
+		flattened[8], flattened[9], flattened[10] = nil, nil, nil
+		flattened[12], flattened[13] = nil, nil
```
liamsi, line 135: Lose data randomly as in other places, too. (e.g. see how @evan-forbes used `generateMissingData`).

extendeddatacrossword_test.go
adlerjohn: Fixed in 8da717f and 01f2eb3. The numbers aren't the best!

```
BenchmarkRepair/Repairing_16x16_ODS_using_RSGF8-4           	     291	   4112360 ns/op
BenchmarkRepair/Repairing_16x16_ODS_using_LeopardFF8-4      	     259	   4406930 ns/op
BenchmarkRepair/Repairing_16x16_ODS_using_LeopardFF16-4     	     266	   4402912 ns/op
BenchmarkRepair/Repairing_32x32_ODS_using_RSGF8-4           	      63	  18601199 ns/op
BenchmarkRepair/Repairing_32x32_ODS_using_LeopardFF8-4      	      63	  18681030 ns/op
BenchmarkRepair/Repairing_32x32_ODS_using_LeopardFF16-4     	      63	  18583115 ns/op
BenchmarkRepair/Repairing_64x64_ODS_using_RSGF8-4           	      14	  78674819 ns/op
BenchmarkRepair/Repairing_64x64_ODS_using_LeopardFF8-4      	      16	  71389948 ns/op
BenchmarkRepair/Repairing_64x64_ODS_using_LeopardFF16-4     	      15	  71138939 ns/op
BenchmarkRepair/Repairing_128x128_ODS_using_RSGF8-4         	       3	 379398435 ns/op
BenchmarkRepair/Repairing_128x128_ODS_using_LeopardFF8-4    	       4	 282697386 ns/op
BenchmarkRepair/Repairing_128x128_ODS_using_LeopardFF16-4   	       4	 293993599 ns/op
```

adlerjohn: Side note: this seems to error when I run with a 256 x 256 square (w/ 256-byte shares). Is that expected? I'm assuming some of the codecs have limits. cc @evan-forbes @liamsi

liamsi: Yeah, the FF8 versions have relatively low limits. You can look them up in the particular libraries.

> The numbers aren't the best!

Happy to do some profiling to understand where the most resources are spent. Remember that we need to add back the last good square which won't help with the numbers.

adlerjohn: I think I can do a bit of profiling as I explore optimizing in #25. It's on the way 🤷. I expect there's room for significant improvement here.

extendeddatacrossword_test.go
```diff
@@ -110,3 +112,45 @@ func TestRepairExtendedDataSquare(t *testing.T) {
...
+			flattened := eds.flattened()
+			// Randomly remove half the shares
+			for i := 0; i < i*i*2; {
+				ind := rand.Intn(i)
+				if len(flattened[ind]) == 0 {
+					continue
+				}
+				flattened[ind] = []byte{}
+				i++
+			}
```
evan-forbes, line 136: It's not too important I don't think, but this doesn't guaranteed delete half the data, it could delete less.

adlerjohn: I agree it's not too important if it doesn't exactly delete half the shares (so long as it's not more than half). But why would it delete less?

evan-forbes: whoops, you're right and I'm blind

adlerjohn: I had actually moved the `i++` into the for loop, then realized that, no, it does in fact belong where it is. 😂

liamsi (approved):
👏🏼

extendeddatacrossword_test.go
liamsi: >  I expect there's room for significant improvement here.

Yeah, I think there is lots of low-hanging fruits here. Looking forward to the optimizations. Wish I've added the benchmarks before #33 already :-D

evan-forbes (approved):
this will be super useful! LGTM :+1:

extendeddatacrossword_test.go
liamsi: As a side note, most oif the time seems to be spent on the pre repair sanity checks:
image

PR #38, opened 2021-04-06
Remove arm64 from test CI

adlerjohn:
GitHub Actions don't have docker containers for ARM architectures. Remove `arm64` from test targets (but still build since we can cross-compile).


liamsi (approved):

liamsi:
Can't run arm64 on linux

PR #41, opened 2021-04-12
Benchmark repair with 1/4 instead of 1/2 of shares

adlerjohn:
Remove 3/4 of shares for repair benchmark to test with the minimum number of shares needed.


liamsi (approved):
Makes sense and LGTM!

Issue #11, opened 2021-01-11
Improve documentation


liamsi:
Currently, some of the exported methods/structs are not properly documented.

Ideally, the readme would contain slightly more information about the repository and how it's meant to be used. The [SMT](https://github.com/lazyledger/smt/) repo's readme is a good example. A runnable [example test](https://blog.golang.org/examples) would also be useful.

PR #42, opened 2021-04-12
Add mem benchmark command to readme.

adlerjohn:
Benchmarking allocations will be important with upcoming optimization attempts, so add `-benchmem` to benchmark command in readme to be clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,5 +13,5 @@ Experimental software. Use at your own risk!
 Run benchmarks
 
 ```sh
-go test -tags leopard -bench=.
+go test -tags leopard -benchmem -bench=.
 ```


liamsi (approved):
So in general, I don't think running benchmarks should be the only thing mentioned in the readme under the usage seciton. Maybe they should not even be mentioned at all.

Instead, there should be more context and a simple example; similar to the SMT / NMT repos:
 - https://github.com/lazyledger/smt
 - https://github.com/lazyledger/nmt

adlerjohn:
I agree around readme documentation overall cf. #11. This PR is purely to document benchmarking _somewhere_ clearly since we'll be running them a lot in the not too distant future. If once things are stable it's no longer relevant it can be removed.

PR #43, opened 2021-04-12
Fix repair benchmark

adlerjohn:
Fix repair to properly drop shares. It now drops exactly half from each row, to guarantee that the square can be repaired. @evan-forbes noted that there are corner cases where the square isn't repairable with <1/2 the EDS's shares (e.g. if only the first rows in their entirety are present).

### DIFF
--- a/extendeddatacrossword_test.go
+++ b/extendeddatacrossword_test.go
@@ -124,15 +124,16 @@ func BenchmarkRepair(b *testing.B) {
 				b.Error(err)
 			}
 
-			flattened := eds.flattened()
-			// Randomly remove 3/4 of the shares
-			for j := 0; j < i*i*3; {
-				ind := rand.Intn(j + 1)
-				if flattened[ind] == nil {
-					continue
+			// Randomly remove 1/2 of the shares of each row
+			for r := 0; r < i*2; {
+				for c := 0; c < i; {
+					ind := rand.Intn(i + 1)
+					if eds.Cell(uint(r), uint(ind)) == nil {
+						continue
+					}
+					eds.setCell(uint(r), uint(ind), nil)
+					c++
 				}
-				flattened[ind] = nil
-				j++
 			}
 
 			b.Run(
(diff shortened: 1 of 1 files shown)


extendeddatacrossword_test.go
```diff
@@ -126,13 +126,13 @@ func BenchmarkRepair(b *testing.B) {
...
 			flattened := eds.flattened()
 			// Randomly remove 3/4 of the shares
-			for i := 0; i < i*i*3; {
-				ind := rand.Intn(i)
+			for j := 0; j < i*i*3; {
+				ind := rand.Intn(j)
 				if len(flattened[ind]) == 0 {
 					continue
 				}
 				flattened[ind] = []byte{}
```
liamsi, line 134:
```suggestion
				flattened[ind] = nil
```

extendeddatacrossword_test.go
```diff
@@ -125,21 +125,24 @@ func BenchmarkRepair(b *testing.B) {
 			}
 
 			flattened := eds.flattened()
-			// Randomly remove 3/4 of the shares
-			for i := 0; i < i*i*3; {
-				ind := rand.Intn(i)
-				if len(flattened[ind]) == 0 {
-					continue
+			// Randomly remove 1/2 of the shares of each row
```
liamsi, line 128: I still think that we should also benchmark losing more data as it presents the worst-case scenario for repairing.

adlerjohn: The issue about removing more than 1/2 is that we're not guaranteed to be able to recover the whole square (i.e. it'll require more manual tweaks). The difference between recovering from 3/4 loss vs 1/2 loss should be a constant factor, so it _should_ irrelevant for benchmarking the relative changes optimizations bring.

liamsi (approved):
👍🏼

extendeddatacrossword_test.go
liamsi: > so it should irrelevant for benchmarking 

Yeah, would be cool to validate this with a benchmark. But you are probably right and it's definitely orthogonal to this PR.

PR #44, opened 2021-04-12
Return uncorrupted EDS from repair, and optimize

### LINKED ISSUE #25: add back LastGoodSquare to errors

adlerjohn:
Fixes #25.

Optimizes repair logic to do work on a so-far correct EDS struct, and only modify it when modifications are guaranteed to not corrupt.

Note that the original implementation did a quadratic amount of work (deep copy of the entire square) each iteration, for a quadratic number of iterations (i.e. `O(n^4)`). This implementation does linear work (process a single row or column) each iteration, for a quadratic number of iterations (i.e `O(n^3)`). To my understanding, this is order-optimal.

## Initial Results

Bench with `edsBackup` re-added as before #24 against `master` branch as of start of this PR:
```
goos: linux
goarch: amd64
pkg: github.com/lazyledger/rsmt2d
cpu: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
BenchmarkRepair/Repairing_16x16_ODS_using_RSGF8-4         	     261	   4228503 ns/op	 3644160 B/op	   11333 allocs/op
BenchmarkRepair/Repairing_32x32_ODS_using_RSGF8-4         	      61	  19182579 ns/op	16971170 B/op	   42245 allocs/op
BenchmarkRepair/Repairing_64x64_ODS_using_RSGF8-4         	      14	  79352834 ns/op	61468033 B/op	  159749 allocs/op
BenchmarkRepair/Repairing_128x128_ODS_using_RSGF8-4       	       1	1495244230 ns/op	2933171416 B/op	 1094336 allocs/op
```

Bench without `edsBackup` logic (`master` branch as of the start of this PR):
```
goos: linux
goarch: amd64
pkg: github.com/lazyledger/rsmt2d
cpu: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
BenchmarkRepair/Repairing_16x16_ODS_using_RSGF8-4         	     277	   4127613 ns/op	 3644169 B/op	   11333 allocs/op
BenchmarkRepair/Repairing_32x32_ODS_using_RSGF8-4         	      61	  19092464 ns/op	16971164 B/op	   42245 allocs/op
BenchmarkRepair/Repairing_64x64_ODS_using_RSGF8-4         	      14	  78191005 ns/op	61468040 B/op	  159749 allocs/op
BenchmarkRepair/Repairing_128x128_ODS_using_RSGF8-4       	       3	 381096647 ns/op	284551544 B/op	  620037 allocs/op
```

We see that removing the deep copy of the EDS at each iteration makes a huge difference at the largest square size. Minor-to-no difference at smaller square sizes. This is mostly within expectations.

Note that the current maximum ODS is 64x64, not 128x128, though that limit is expected to be raised either before mainnet or not too long thereafter.

## Optimized Results

Bench with optimized EDS backup logic:

```
goos: linux
goarch: amd64
pkg: github.com/lazyledger/rsmt2d
cpu: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
BenchmarkRepair/Repairing_16x16_ODS_using_RSGF8-4         	     273	   4292978 ns/op	 3644160 B/op	   11333 allocs/op
BenchmarkRepair/Repairing_32x32_ODS_using_RSGF8-4         	      61	  19140366 ns/op	16971166 B/op	   42245 allocs/op
BenchmarkRepair/Repairing_64x64_ODS_using_RSGF8-4         	      14	  77965243 ns/op	61468058 B/op	  159749 allocs/op
BenchmarkRepair/Repairing_128x128_ODS_using_RSGF8-4       	       3	 376856007 ns/op	284551544 B/op	  620037 allocs/op
```

We see that the optimized version performs the same as the version without EDS deep copies.

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -85,6 +85,7 @@ func RepairExtendedDataSquare(
 }
 
 func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][]byte, bitMask bitMatrix) error {
+	// TODO(john): re-add eds
 	// Keep repeating until the square is solved
 	solved := false
 	for {
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -272,21 +272,21 @@ func (eds *ExtendedDataSquare) prerepairSanityCheck(rowRoots [][]byte, columnRoo
 		}
 
 		if rowIsComplete {
-			shares, err = Encode(eds.rowSlice(i, 0, eds.originalDataWidth), eds.codec)
+			parityShares, err := Encode(eds.rowSlice(i, 0, eds.originalDataWidth), eds.codec)
 			if err != nil {
 				return err
 			}
-			if !bytes.Equal(flattenChunks(shares), flattenChunks(eds.rowSlice(i, eds.originalDataWidth, eds.originalDataWidth))) {
+			if !bytes.Equal(flattenChunks(parityShares), flattenChunks(eds.rowSlice(i, eds.originalDataWidth, eds.originalDataWidth))) {
 				return &ErrByzantineRow{i}
 			}
 		}
 
 		if colIsComplete {
-			shares, err = Encode(eds.columnSlice(0, i, eds.originalDataWidth), eds.codec)
+			parityShares, err := Encode(eds.columnSlice(0, i, eds.originalDataWidth), eds.codec)
 			if err != nil {
 				return err
 			}
-			if !bytes.Equal(flattenChunks(shares), flattenChunks(eds.columnSlice(eds.originalDataWidth, i, eds.originalDataWidth))) {
+			if !bytes.Equal(flattenChunks(parityShares), flattenChunks(eds.columnSlice(eds.originalDataWidth, i, eds.originalDataWidth))) {
 				return &ErrByzantineColumn{i}
 			}
 		}
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -18,8 +18,7 @@ var ErrUnrepairableDataSquare = errors.New("failed to solve data square")
 // ErrByzantineRow is thrown when there is a repaired row does not match the expected row Merkle root.
 type ErrByzantineRow struct {
 	RowNumber uint     // Row index
-	Root      []byte   // Merkle root of repaired row
-	Shares    [][]byte // Repaired row shares
+	Shares    [][]byte // Pre-repaired row shares
 }
 
 func (e *ErrByzantineRow) Error() string {
@@ -29,8 +28,7 @@ func (e *ErrByzantineRow) Error() string {
 // ErrByzantineColumn is thrown when there is a repaired column does not match the expected column Merkle root.
 type ErrByzantineColumn struct {
 	ColumnNumber uint     // Column index
-	Root         []byte   // Merkle root of repaired column
-	Shares       [][]byte // Repaired column shares
+	Shares       [][]byte // Pre-repaired column shares
 }
 
 func (e *ErrByzantineColumn) Error() string {
@@ -49,8 +47,9 @@ func (e *ErrByzantineColumn) Error() string {
 //
 // The EDS is modified in-place. If repairing is successful, the EDS will be
 // complete. If repairing is unsuccessful, the EDS will be the most-repaired
-// prior to the Byzantine row or column being repaired, and the repaired
-// Byzantine row or column is returned in the error.
+// prior to the Byzantine row or column being repaired, and the Byzantine row
+// or column prior to repair is returned in the error with missing shares as
+// nil.
 func RepairExtendedDataSquare(
 	rowRoots [][]byte,
 	columnRoots [][]byte,
@@ -178,7 +177,7 @@ func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][
 					}
 
 					// Check that rebuilt shares matches appropriate root
-					err = eds.verifyAgainstRoots(rowRoots, columnRoots, mode, uint(i), rebuiltShares)
+					err = eds.verifyAgainstRoots(rowRoots, columnRoots, mode, uint(i), bitMask, rebuiltShares)
 					if err != nil {
 						return err
 					}
@@ -189,7 +188,7 @@ func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][
 						case row:
 							if !bitMask.Get(i, j) &&
 								bitMask.ColumnIsOne(j) {
-								err := eds.verifyAgainstRoots(rowRoots, columnRoots, column, uint(j), rebuiltShares)
+								err := eds.verifyAgainstRoots(rowRoots, columnRoots, column, uint(j), bitMask, rebuiltShares)
 								if err != nil {
 									return err
 								}
@@ -198,7 +197,7 @@ func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][
 						case column:
 							if !bitMask.Get(j, i) &&
 								bitMask.RowIsOne(j) {
-								err := eds.verifyAgainstRoots(rowRoots, columnRoots, row, uint(j), rebuiltShares)
+								err := eds.verifyAgainstRoots(rowRoots, columnRoots, row, uint(j), bitMask, rebuiltShares)
 								if err != nil {
 									return err
 								}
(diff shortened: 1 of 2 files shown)


adlerjohn:
For posterity the leopard tag doesn't work with this branch because of the following bug.

The `Decode` (`Encode`) function for Leopard. The function with `RSGF8` only returns the original shares (parity shares), while the `Decode` (`Encode`) function with `Leopard{FF8,FF16}` returns both the original and parity shares. This was a silent bug because only the first `originalDataWidth` elements of each returned slice were used, so the difference in behavior happened to be ignored. But when doing other manipulation (e.g. `append` the difference in behavior is noticed).

a2a6f15 implements a workaround for half the issue (the decoding half), but can't fully workaround with encoding half.

liamsi:
> This was a silent bug because only the first originalDataWidth elements of each returned slice were used, so the difference in behavior happened to be ignored. But when doing other manipulation (e.g. append the difference in behavior is noticed).

Does that mean everything works as expected unless you go and modify the returned slices? Couldn't a workaround be to just copy the desired part of the slice before e.g. appending to it? That's what I would do to fix this in go-leopard before diving into the c++ code of the actual leopard library again.

adlerjohn:
@liamsi did a full workaround, so things work now with Leopard.

extendeddatacrossword.go
```diff
@@ -85,6 +85,7 @@ func RepairExtendedDataSquare(
 }
 
 func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][]byte, bitMask bitMatrix) error {
+	// TODO(john): re-add eds
```
liamsi, line 88: Is this outdated?

liamsi (commented):
> We see that the optimized version performs the same as the version without EDS deep copies.

Can we use [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) for these kinds of comparisons moving forward?

extendeddatacrossword.go
```diff
@@ -191,44 +233,30 @@ func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][
...
-		if mode == row {
-			eds.setCell(rowOrColIdx, eds.originalDataWidth+uint(p), s)
-		} else if mode == column {
-			eds.setCell(eds.originalDataWidth+uint(p), rowOrColIdx, s)
+	root := tree.Root()
+
+	switch mode {
+	case row:
+		if !bytes.Equal(root, rowRoots[i]) {
+			return nil, &ErrByzantineRow{i}
```
liamsi, line 247: I thought the whole point of this PR to add back the eds in these errors?

extendeddatacrossword.go
adlerjohn: Yes!

adlerjohn: Fixed in 9a6444a.

extendeddatacrossword.go
adlerjohn: No, this PR implements https://github.com/lazyledger/rsmt2d/issues/25#issue-849746001

> Basically, instead of corrupting the EDS then doing the check, do the check first then set the EDS in a way that's guaranteed correct. The check would have to compute a root based on existing leaves and a new set of repaired leaves, but that's just copy-pasting this logic.

There's no reason to return the uncorrupted EDS, since it's modified in-place and available to the caller. If anything, what should be added to the error is the bad row/column shares.

liamsi: OK, that makes sense but is non-obvious. Can you please add a test that explains how the errors are meant to be used?

adlerjohn: Fixed to return root and shares in 7c32130 and f8d10e6.

extendeddatacrossword.go
```diff
@@ -324,3 +309,11 @@ func noMissingData(input [][]byte) bool {
...
 	return true
 }
+
+func (eds *ExtendedDataSquare) computeSharesRoot(shares [][]byte, i uint) []byte {
+	tree := eds.createTreeFn()
+	for cell, d := range shares {
+		tree.Push(d, SquareIndex{Cell: uint(cell), Axis: i})
+	}
+	return tree.Root()
+}
```
liamsi, line 319: 👍🏼 This is a bit orthogonal remark, but I always wondered if the API of the underlying tree should not take a list of leaves directly and return the root instead of pushing these in a loop instead. I think everywhere we are using the tree, we always have all leaves at hand at once anyways.

adlerjohn: Yeah probably! Intuitively, I don't think this is a huge overhead though (worst-case amortized linear cost, so order-optimal), so maybe lower-priority.

extendeddatacrossword.go
adlerjohn: I improved the documentation of what the errors includes in 7d8f5e0, is that sufficient @liamsi? A test probably wouldn't be particularly helpful, because it's only used to construct a fraud proof.

liamsi: These comments are helpful indeed. 

> because it's only used to construct a fraud proof.

Yeah, couldn't the test construct and validate a fraud-proof? it would be the best and most useful documentation IMO.

adlerjohn: > Yeah, couldn't the test construct and validate a fraud-proof?

Maybe, but I feel like that's outside the scope of this PR. In fact, it's probably outside the scope of this entire library! The exact structure of a fraud proofs is LazyLedger-specific.

IMO I should probably just add even more documentation for expected pre- and post-conditions of the exported functions.

adlerjohn: Expanded doc in 1f987bb that should make it much more obvious what's happening.

liamsi: > In fact, it's probably outside the scope of this entire library! 

I disagree and mostly because these errors and their fields are part of the public API and tests are there to test and also document that public API.

> Expanded doc in 1f987bb that should make it much more obvious what's happening.

That is really helpful too though.

liamsi: > The exact structure of a fraud proof is LazyLedger-specific.

Maybe the test can then be general enough to show what you can prove with the returned data? More or less a poor mans fraud proof?

liamsi: Before I've commented, it wasn't even clear that the bad row/column shares need to be part of that error. Via a test, we, well, test the assumption of how this can be used.

adlerjohn: Fixed in 13de53c. Instead of returning the bad repaired shares, the error returns the pre-repair good shares, with missing shares as `nil`. It's then used in a pseudo fraud proof in the tests. Good call on writing a test, as it identified exactly what needs to be returned.

Note that the real fraud proof would include a Merkle proof for each non-`nil` share that must then be verified, however doing this was not functionality that this library implemented previously. As such, I think it should be done in a future PR. Also, it's non-trivial, so breaking up the work into multiple PRs will make changes more digestible.

extendeddatacrossword.go
```diff
@@ -94,87 +109,128 @@ func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][
...
+				switch mode {
+				case row:
 					isIncomplete = !bitMask.RowIsOne(i)
 					isExtendedPartIncomplete = !bitMask.RowRangeIsOne(i, int(eds.originalDataWidth), int(eds.width))
-				} else if mode == column {
+				case column:
 					isIncomplete = !bitMask.ColumnIsOne(i)
 					isExtendedPartIncomplete = !bitMask.ColRangeIsOne(i, int(eds.originalDataWidth), int(eds.width))
+				default:
+					panic(fmt.Sprintf("invalid mode %d", mode))
```
liamsi, line 122: IMO, we should simplify this by getting rid of the `mode`. But orthogonal to this PR.

extendeddatacrossword.go
```diff
@@ -94,87 +109,128 @@ func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][
...
-								eds.setCell(uint(p), uint(i), s)
-							}
-						}
-						if isExtendedPartIncomplete {
-							err := eds.rebuildExtendedPart(mode, uint(i))
-							if err != nil {
-								return err
-							}
-						}
+						continue
```
liamsi, line 154: ~~Can't this cause an endless loop now?~~ NVM, this only skips the inner loop so it should be fine. Still difficult to follow the for-loop imo (but also orthogonal).

extendeddatacrossword.go
```diff
@@ -94,87 +109,128 @@ func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][
...
+						startIndex := len(rebuiltExtendedShares) - int(eds.originalDataWidth)
+						rebuiltShares = append(
+							rebuiltShares[0:eds.originalDataWidth],
+							rebuiltExtendedShares[startIndex:]...,
+						)
+					} else {
+						// Otherwise copy them from the EDS.
+						startIndex := len(shares) - int(eds.originalDataWidth)
+						rebuiltShares = append(
+							rebuiltShares[0:eds.originalDataWidth],
```
liamsi, line 174: Can you still open an issue in the go-leopard library about the diff in the API/returned results.

extendeddatacrossword.go
```diff
@@ -94,87 +109,128 @@ func (eds *ExtendedDataSquare) solveCrossword(rowRoots [][]byte, columnRoots [][
...
+					// Check that newly completed orthogonal vectors match their new merkle roots
+					for j := 0; j < int(eds.width); j++ {
+						switch mode {
+						case row:
+							if !bitMask.Get(i, j) &&
+								bitMask.ColumnIsOne(j) {
+								err := eds.verifyAgainstRoots(rowRoots, columnRoots, column, uint(j), bitMask, rebuiltShares)
+								if err != nil {
+									return err
 								}
```
liamsi, line 194: We should aim for full coverage on the non-happy path branches too. (also orthogonal to this PR)

liamsi (approved):
Great work 👏🏼

extendeddatacrossword.go
adlerjohn: Opened https://github.com/lazyledger/go-leopard/issues/7

PR #45, opened 2021-04-14
Test with leopard in CI

adlerjohn:
Run CI tests with `-tags leopard`.

Note: had to remove `arm64` as a build target since cgo doesn't work with cross-compiling.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
           go-version: 1.15
       - name: test & coverage report creation
         run: |
-          GOARCH=${{ matrix.goarch }} go test -mod=readonly -timeout 8m -race -coverprofile=coverage.txt -covermode=atomic
+          GOARCH=${{ matrix.goarch }} go test -tags leopard -mod=readonly -timeout 8m -race -coverprofile=coverage.txt -covermode=atomic
         if: env.GIT_DIFF
       - uses: codecov/codecov-action@v1.0.15
         with:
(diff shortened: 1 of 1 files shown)


liamsi:
We don't see if this actually worked bc there is no diff in any go file. Do you mind changing CI to temporarily run even if there was no diff? I'm a bit surprised that only the build tag is necessary. On the other hand, we'd already see any issues regarding leopard during build-time I think.

adlerjohn:
Hmm. I forget, were there any steps for installing `libleopard`?

liamsi:
OK, now we see that this didn't work: 

```
# github.com/lazyledger/go-leopard/leopard
/usr/bin/ld: cannot find -llibleopard
/usr/bin/ld: cannot find -llibleopard
/usr/bin/ld: cannot find -llibleopard
/usr/bin/ld: cannot find -llibleopard
collect2: error: ld returned 1 exit status
FAIL	github.com/lazyledger/rsmt2d [build failed]
```

Currently, you'd need to add a few lines to actually build leopard and install it properly. Maybe this should be done in a Makefile. That way it would be a single line in the CI config and it could be re-used for other purposes, too.

liamsi:
https://github.com/lazyledger/go-leopard/blob/50ec4b3fab41cd3af5adc88a3447f265acf59352/Makefile#L14-L20

liamsi:
Should we merge this in first? I mean before #44 ?

adlerjohn:
> Should we merge this in first?

Probably not, we'd want to get the CI actually installing `libleopard` and stuff first before testing with it so that it actually works. Converted to draft while I fix this. #44 is independent.

liamsi (commented):
lgtm 👍🏼

liamsi (approved):
👍🏼

PR #46, opened 2021-04-15
Refactor Codec API to allow for thread safe operation of the infectious RSGF8

evan-forbes:
This PR refactors the `Codec` api to a more typical go interface, while still maintaining separation of the leopard library using the leopard build tag. This Allows for the RSGF8 codec to be used in a thread safe fashion.

### DIFF
--- a/codecs.go
+++ b/codecs.go
@@ -30,12 +30,12 @@ func NewLeoRSFF16Codec() Codec {
 	if codec, has := codecs[LeopardFF16]; has {
 		return codec
 	}
-	panic("cannot use codec LeopardFF16 without 'leopard' build tag")
+	panic("cannot use codec LeopardFF16 without the 'leopard' build tag")
 }
 
 func NewLeoRSFF8Codec() Codec {
 	if codec, has := codecs[LeopardFF8]; has {
 		return codec
 	}
-	panic("cannot use codec LeopardFF8 without 'leopard' build tag")
+	panic("cannot use codec LeopardFF8 without the 'leopard' build tag")
 }
--- a/codecs.go
+++ b/codecs.go
@@ -11,7 +11,6 @@ const (
 type Codec interface {
 	Encode(data [][]byte) ([][]byte, error)
 	Decode(data [][]byte) ([][]byte, error)
-	New() Codec
 	// maxChunks returns the max. number of chunks each code supports in a 2D square.
 	maxChunks() int
 }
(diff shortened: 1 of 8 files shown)


extendeddatasquare.go
```diff
@@ -10,26 +10,21 @@ import (
 type ExtendedDataSquare struct {
 	*dataSquare
 	originalDataWidth uint
-	codec             CodecType
 }
```
evan-forbes, line 13: I didn't actually have to remove this codec as a field, but removing it made the api consistent. Every time a codec is needed, the user is required to provide one. This forces the user to consciously choose an already used codec, or a new instance of that codec.

codecs.go
```diff
@@ -1,63 +1,41 @@
...
 type Codec interface {
-	encode(data [][]byte) ([][]byte, error)
-	decode(data [][]byte) ([][]byte, error)
-	codecType() CodecType
+	Encode(data [][]byte) ([][]byte, error)
+	Decode(data [][]byte) ([][]byte, error)
+	New() Codec
 	// maxChunks returns the max. number of chunks each code supports in a 2D square.
 	maxChunks() int
 }
```
evan-forbes, line 17: the extra `New` method is also not needed, but it will useful in the future should we be using mutlithreaded encoding/repairing

liamsi: Just out of curiosity: Why are the `Encode`/`Decode` methods exported now? 

> the extra New method is also not needed, but it will useful in the future should we be using mutlithreaded encoding/repairing

Can we introduce it when we need it instead?

evan-forbes: The `Encode` and `Decode` methods are exported because the functions with the same names have been removed.

codecs.go
```diff
@@ -1,63 +1,40 @@
...
 		panic(fmt.Sprintf("%v already registered", codec))
 	}
 	codecs[ct] = codec
 }
 
-func Encode(data [][]byte, codec CodecType) ([][]byte, error) {
-	if codec, ok := codecs[codec]; !ok {
-		return nil, errors.New("invalid codec")
-	} else {
-		return codec.encode(data)
```
liamsi, line 53: I see, that's why the Encode/Decode methods need to be exported on the interface now. Wich makes more sense too.

codecs.go
liamsi: Yeah, just saw that https://github.com/lazyledger/rsmt2d/pull/46#discussion_r614427643
Thanks for clarifying.

liamsi (approved):
Nice work! I think we can soon close #12

adlerjohn (commented):
Nice work

PR #47, opened 2021-04-16
Print codec name instead of codec for repair bench

adlerjohn:
Fix printing bug introduced in #46.


evan-forbes (approved):
lol whoops. Thanks! I should have ran the benchmarks!!

liamsi (approved):

PR #49, opened 2021-04-17
Fix repair bench to remove half from each row.

adlerjohn:
Use `i * 2` as width of each row instead of `i` (`i` is ODS width).

### DIFF
--- a/extendeddatacrossword_test.go
+++ b/extendeddatacrossword_test.go
@@ -153,18 +153,6 @@ func BenchmarkRepair(b *testing.B) {
 			}
 
 			extendedDataWidth := originalDataWidth * 2
-			flattened := eds.flattened()
-			// Randomly remove 1/2 of the shares of each row
-			for r := 0; r < extendedDataWidth; r++ {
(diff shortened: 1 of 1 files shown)


extendeddatacrossword_test.go
```diff
@@ -156,11 +156,11 @@ func BenchmarkRepair(b *testing.B) {
 			// Randomly remove 1/2 of the shares of each row
 			for r := 0; r < i*2; r++ {
 				for c := 0; c < i; {
-					ind := rand.Intn(i + 1)
-					if flattened[r*i+ind] == nil {
+					ind := rand.Intn(i * 2)
```
liamsi, line 159: I'd recommend renaming `i` to something more meaningful, e.g. `origDataSize` and used `eds.Width()` or a new variable instead of `i*2` (e.g. extended width).

adlerjohn: Not a bad idea.

adlerjohn: Fixed in b0c81c7.

liamsi (approved):
👍🏼

PR #50, opened 2021-04-19
Fix repair bench to re-initialize shares

adlerjohn:
The repair benchmark was re-using repaired shares, resulting in only a single run of the actual repairing following by `b.N` runs of just checking that the repaired shares are correct. Fix to actually re-initialize the missing shares on each run.

Depends on #49.

### DIFF
--- a/extendeddatacrossword_test.go
+++ b/extendeddatacrossword_test.go
@@ -153,18 +153,6 @@ func BenchmarkRepair(b *testing.B) {
 			}
 
 			extendedDataWidth := originalDataWidth * 2
-			flattened := eds.flattened()
-			// Randomly remove 1/2 of the shares of each row
-			for r := 0; r < extendedDataWidth; r++ {
-				for c := 0; c < originalDataWidth; {
-					ind := rand.Intn(extendedDataWidth)
-					if flattened[r*extendedDataWidth+ind] == nil {
-						continue
-					}
-					flattened[r*extendedDataWidth+ind] = nil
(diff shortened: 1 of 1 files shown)


extendeddatacrossword_test.go
```diff
@@ -175,6 +163,23 @@ func BenchmarkRepair(b *testing.B) {
...
+								if flattened[r*extendedDataWidth+ind] == nil {
+									continue
+								}
+								flattened[r*extendedDataWidth+ind] = nil
+								c++
+							}
+						}
+
+						b.StartTimer()
+
```
liamsi, line 182: IMO, ideally we should use table-driven tests here too. But this can be enhanced in a separate PR.this

liamsi (approved):
Curious how much that changes the numbers in https://github.com/lazyledger/rsmt2d/pull/48

LGTM

extendeddatacrossword_test.go
```diff
@@ -175,6 +163,23 @@ func BenchmarkRepair(b *testing.B) {
...
 						_, err := RepairExtendedDataSquare(
```
liamsi, line 183: Shouldn't this actually set the eds back to fully repaired though with each iteration? I see that the benchmarks work  without it but I'm not sure I understand why tbh.

liamsi: Actually, NVM. It's only the `flattened` slice that gets modified on each iteration and not the EDS...

PR #55, opened 2021-04-21
Refactor `solveCrossword` main loop to separate row and col processing

### LINKED ISSUE #54: Proposal: refactor main for-loop in solveCrossword
The different "modes" are detrimental to readability and do not make much sense. With the last change in #52 it becomes even more apparent that we should simply refactor this in a way that first handles rows and then columns (in two different methods/functions to reduce that nesting). Instead of an added for-loop and then checking each time if we are in row or in column "mode":

https://github.com/lazyledger/rsmt2d/blob/d3add80beac48d64a077100585d4156b26a1a504/extendeddatacrossword.go#L212-L223
https://github.com/lazyledger/rsmt2d/blob/d3add80beac48d64a077100585d4156b26a1a504/extendeddatacrossword.go#L227-L234
https://github.com/lazyledger/rsmt2d/blob/d3add80beac48d64a077100585d4156b26a1a504/extendeddatacrossword.go#L187-L209

And many more places.

I tried this in this branch a while ago (time really flies): https://github.com/lazyledger/rsmt2d/commits/ismail/wip_solveCrossword_readability_refactoring particularly in this commit: https://github.com/lazyledger/rsmt2d/commit/d88f1ab4ad4a5df02ebde8a3ef9906ecd0ac25ff

Ideally, the result would be that the main loop is simple to grasp when looking at the code (without reading any comments like "Keep repeating until the square is solved" necessarily):

https://github.com/lazyledger/rsmt2d/blob/d3add80beac48d64a077100585d4156b26a1a504/extendeddatacrossword.go#L102-L248

adlerjohn:
Fixes #54.

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -139,7 +139,6 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(r int, rowRoots [][]byte, colum
 		return true, false, nil
 	}
 
-	isExtendedPartIncomplete := !bitMask.RowRangeIsOne(r, int(eds.originalDataWidth), int(eds.width))
 	// Prepare shares
 	shares := make([][]byte, eds.width)
 	for c := 0; c < int(eds.width); c++ {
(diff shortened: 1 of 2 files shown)


extendeddatacrossword.go
```diff
@@ -101,177 +101,208 @@ func RepairExtendedDataSquare(
...
+			if err != nil {
+				return err
+			}
+			solvedCol, progressMadeCol, err := eds.solveCrosswordCol(i, rowRoots, columnRoots, bitMask, codec)
+			if err != nil {
+				return err
 			}
+
+			solved = solved && solvedRow && solvedCol
+			progressMade = progressMade || progressMadeRow || progressMadeCol
```
liamsi, line 122: This is so much better 👍

liamsi (commented):
This looks great 👍 I'll do another pass when I'm back on a larger screen (instead of phone).

extendeddatacrossword.go
```diff
@@ -101,177 +101,208 @@ func RepairExtendedDataSquare(
...
-						switch mode {
-						case row:
-							eds.setCell(uint(i), uint(p), s)
-						case column:
-							eds.setCell(uint(p), uint(i), s)
-						default:
-							panic(fmt.Sprintf("invalid mode %d", mode))
-						}
-					}
-				}
```
liamsi, line 236: I had to zoom out in the browser to see and appreciate all these red lines :-P :-D 

image

extendeddatacrossword.go
```diff
@@ -101,177 +101,208 @@ func RepairExtendedDataSquare(
...
+			solvedRow, progressMadeRow, err := eds.solveCrosswordRow(i, rowRoots, columnRoots, bitMask, codec)
+			if err != nil {
+				return err
+			}
```
liamsi, line 115: We should open an issue about covering all the non-happy paths if it doesn't exist already.

liamsi (approved):
This is great!

PR #56, opened 2021-04-23
Improve documentation

### LINKED ISSUE #11: Improve documentation

adlerjohn:
Fixes #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,71 @@ Go implementation of [two dimensional Reed-Solomon Merkle tree data availability
 [![Codecov](https://img.shields.io/codecov/c/github/lazyledger/rsmt2d)](https://app.codecov.io/gh/lazyledger/rsmt2d)
 [![GoDoc](https://godoc.org/github.com/lazyledger/rsmt2d?status.svg)](https://godoc.org/github.com/lazyledger/rsmt2d)
 
-## Usage
+## Example
+
+```go
+package main
+
+import (
+    "bytes"
+
+    "github.com/lazyledger/rsmt2d"
+)
+
+func main() {
+    // Size of each share, in bytes
+    bufferSize := 64
+    // Init new codec
+    codec := rsmt2d.NewLeoRSFF16Codec()
+
+    ones := bytes.Repeat([]byte{1}, bufferSize)
+    twos := bytes.Repeat([]byte{2}, bufferSize)
+    threes := bytes.Repeat([]byte{3}, bufferSize)
+    fours := bytes.Repeat([]byte{4}, bufferSize)
+
+    // Compute parity shares
+    eds, _ := rsmt2d.ComputeExtendedDataSquare(
+        [][]byte{
+            ones, twos,
+            threes, fours,
+        },
+        codec,
+        rsmt2d.NewDefaultTree,
+    )
+
+    // Save all shares in flattended form.
+    // Note: slices returned from Row() an Column() are read-only.
+    // If you need to write to them, copy first.
+    flattened := make([][]byte, 0, eds.Width()*eds.Width())
+    for i := uint(0); i < eds.Width(); i++ {
+        startIndex := i * eds.Width()
+        copy(flattened[startIndex:startIndex+eds.Width()], eds.Row(i))
+    }
+
+    // Delete some shares, just enough so that repairing is possible.
+    flattened[0], flattened[2], flattened[3] = nil, nil, nil
+    flattened[4], flattened[5], flattened[6], flattened[7] = nil, nil, nil, nil
+    flattened[8], flattened[9], flattened[10] = nil, nil, nil
+    flattened[12], flattened[13] = nil, nil
+
+    // Repair square.
+    repaired, err := rsmt2d.RepairExtendedDataSquare(
+        eds.RowRoots(),
+        eds.ColumnRoots(),
+        flattened,
+        codec,
+        rsmt2d.NewDefaultTree,
+    )
+    if err != nil {
+        // err contains information to construct a fraud proof
+        // See extendeddatacrossword_test.go
+    }
+    _ = repaired
+}
+
+```
+
+## Building From Source
 
 Run benchmarks
 
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ func main() {
     // Size of each share, in bytes
     bufferSize := 64
     // Init new codec
-    codec := rsmt2d.NewLeoRSFF16Codec()
+    codec := rsmt2d.NewLeoRSFF8Codec()
 
     ones := bytes.Repeat([]byte{1}, bufferSize)
     twos := bytes.Repeat([]byte{2}, bufferSize)
--- a/README.md
+++ b/README.md
@@ -1,12 +1,10 @@
 # rsmt2d
 
-Go implementation of two dimensional Reed-Solomon merkle tree data availability scheme.
+Go implementation of [two dimensional Reed-Solomon Merkle tree data availability scheme](https://arxiv.org/abs/1809.09044).
 
 [![GitHub Workflow Status](https://img.shields.io/github/workflow/status/lazyledger/rsmt2d/Tests)](https://github.com/lazyledger/rsmt2d/actions/workflows/ci.yml)
 [![Codecov](https://img.shields.io/codecov/c/github/lazyledger/rsmt2d)](https://app.codecov.io/gh/lazyledger/rsmt2d)
 [![GoDoc](https://godoc.org/github.com/lazyledger/rsmt2d?status.svg)](https://godoc.org/github.com/lazyledger/rsmt2d)
-
-Experimental software. Use at your own risk!
 
 ## Usage
 
--- a/datasquare.go
+++ b/datasquare.go
@@ -105,6 +105,7 @@ func (ds *dataSquare) rowSlice(x uint, y uint, length uint) [][]byte {
 }
 
 // Row returns the data in a row.
+// Do not modify this slice directly, instead use SetCell.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1,14 +1,76 @@
...
+    "bytes"
+
+    "github.com/lazyledger/rsmt2d"
+)
+
+func main() {
+    // Size of each share, in bytes
+    bufferSize := 64
+    // Init new codec
+    codec := rsmt2d.NewLeoRSFF16Codec()
```
liamsi, line 24: I feel like the default example should not rely on cgo / leopard.

datasquare.go
```diff
@@ -102,6 +105,7 @@ func (ds *dataSquare) rowSlice(x uint, y uint, length uint) [][]byte {
 }
 
 // Row returns the data in a row.
+// Do not modify this slice directly.
 func (ds *dataSquare) Row(x uint) [][]byte {
```
liamsi, line 109: Wouldn't a non-pointer receiver be better in case we don't want consumers to modify the returned slice?
```suggestion
func (ds dataSquare) Row(x uint) [][]byte {
```

datasquare.go
```diff
@@ -128,6 +132,7 @@ func (ds *dataSquare) columnSlice(x uint, y uint, length uint) [][]byte {
 }
 
 // Column returns the data in a column.
+// Do not modify this slice directly.
 func (ds *dataSquare) Column(y uint) [][]byte {
```
liamsi, line 136: Same as above.

README.md
adlerjohn: Good point.

datasquare.go
adlerjohn: But this would copy the entire data square each time one of these methods is called. Wouldn't it better to make the exported methods return a copy, and make new internal methods that return a slice?

liamsi: Fair enough. Note that the type `dataSquare` is unexported itself. But then it is just embedded into an `ExtendedDataSquare`. Hence these methods on an unexported type are still public API (which isn't obvious when looking at the extended data square). Ideally, we would make this more explicit, too.

> Wouldn't it better to make the exported methods return a copy of the slice

Yeah, probably. I guess this is orthogonal for this PR and we can track it in a separate issue.

evan-forbes: I might have messed up running the benchmarks, but removing the pointers for read only methods doesn't seem to make a noticeable difference time wise. I like the exported copy method too. Regardless, I definitely agree that it's orthogonal.

```
name                                               old time/op  new time/op   delta
Encoding/Encoding_128_shares_using_RSGF8-16         269µs ± 0%    259µs ± 0%   ~     (p=1.000 n=1+1)
Encoding/Encoding_128_shares_using_LeopardFF8-16    145µs ± 0%    149µs ± 0%   ~     (p=1.000 n=1+1)
Encoding/Encoding_128_shares_using_LeopardFF16-16   143µs ± 0%    146µs ± 0%   ~     (p=1.000 n=1+1)
Decoding/Decoding_128_shares_using_RSGF8-16        57.4µs ± 0%   55.8µs ± 0%   ~     (p=1.000 n=1+1)
Decoding/Decoding_128_shares_using_LeopardFF8-16    283µs ± 0%    281µs ± 0%   ~     (p=1.000 n=1+1)
Decoding/Decoding_128_shares_using_LeopardFF16-16   288µs ± 0%    275µs ± 0%   ~     (p=1.000 n=1+1)
Roots/Square_Size_32x32-16                         1.73µs ± 0%   2.55µs ± 0%   ~     (p=1.000 n=1+1)
Roots/Square_Size_64x64-16                         3.68µs ± 0%   5.34µs ± 0%   ~     (p=1.000 n=1+1)
Roots/Square_Size_128x128-16                       6.53µs ± 0%   9.70µs ± 0%   ~     (p=1.000 n=1+1)
Roots/Square_Size_256x256-16                       9.49µs ± 0%  12.81µs ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_16x16_ODS_using_RSGF8-16          7.05ms ± 0%   7.25ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_16x16_ODS_using_LeopardFF8-16     9.83ms ± 0%   9.68ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_16x16_ODS_using_LeopardFF16-16    10.1ms ± 0%    9.9ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_32x32_ODS_using_RSGF8-16          33.8ms ± 0%   33.4ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_32x32_ODS_using_LeopardFF8-16     45.2ms ± 0%   43.5ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_32x32_ODS_using_LeopardFF16-16    44.4ms ± 0%   44.8ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_64x64_ODS_using_LeopardFF16-16     121ms ± 0%    118ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_64x64_ODS_using_RSGF8-16           104ms ± 0%    103ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_64x64_ODS_using_LeopardFF8-16      119ms ± 0%    122ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_128x128_ODS_using_RSGF8-16         437ms ± 0%    442ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_128x128_ODS_using_LeopardFF8-16    413ms ± 0%    423ms ± 0%   ~     (p=1.000 n=1+1)
Repair/Repairing_128x128_ODS_using_LeopardFF16-16   407ms ± 0%    411ms ± 0%   ~     (p=1.000 n=1+1)
Extension/RSGF8_size_4-16                          36.6µs ± 0%   37.6µs ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF8_size_4-16                     84.9µs ± 0%   89.7µs ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF16_size_4-16                    85.5µs ± 0%   87.9µs ± 0%   ~     (p=1.000 n=1+1)
Extension/RSGF8_size_8-16                           165µs ± 0%    159µs ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF8_size_8-16                      326µs ± 0%    334µs ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF16_size_8-16                     315µs ± 0%    334µs ± 0%   ~     (p=1.000 n=1+1)
Extension/RSGF8_size_16-16                          902µs ± 0%    915µs ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF8_size_16-16                    1.34ms ± 0%   1.39ms ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF16_size_16-16                   1.33ms ± 0%   1.29ms ± 0%   ~     (p=1.000 n=1+1)
Extension/RSGF8_size_32-16                         5.71ms ± 0%   5.80ms ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF8_size_32-16                    6.43ms ± 0%   6.06ms ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF16_size_32-16                   6.25ms ± 0%   6.12ms ± 0%   ~     (p=1.000 n=1+1)
Extension/RSGF8_size_64-16                         23.7ms ± 0%   22.9ms ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF8_size_64-16                    16.5ms ± 0%   16.6ms ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF16_size_64-16                   16.8ms ± 0%   17.1ms ± 0%   ~     (p=1.000 n=1+1)
Extension/RSGF8_size_128-16                         133ms ± 0%    130ms ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF8_size_128-16                   48.7ms ± 0%   49.0ms ± 0%   ~     (p=1.000 n=1+1)
Extension/LeopardFF16_size_128-16                  50.1ms ± 0%   50.6ms ± 0%   ~     (p=1.000 n=1+1)
```

liamsi: Another consideration here is consistency:
> Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used. 

https://golang.org/doc/faq#methods_on_values_or_pointers

Here is what I think: make all exported methods on unexported types unexported too. Think what we want to expose and expose this more explicitly. During that refactor the question of pointer vs values will come up naturally anyways.

adlerjohn: > but removing the pointers for read only methods doesn't seem to make a noticeable difference time wise

Ah, I think that's because it only does a shallow copy, and the majority of the data square is the slices of the shares.

adlerjohn: @liamsi so I'll leave refactoring this to a future PR, to keep this PR contained to just better documentation.

README.md
adlerjohn: Fixed in e500007 (that's a lot of zeroes).

evan-forbes (approved):
two :+1: :+1: from me!

datasquare.go
liamsi: Yes, sounds good to me.

PR #57, opened 2021-05-05
Improve public API

adlerjohn:
Partially addresses #12.

Don't export `dataSquare` methods, and instead only export `ExtendedDataSquare` methods that wrap those. Functionally mostly the same, except `Row` and `Col` now return copies, so the caller can't accidentally write to internal slices in a bad way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,8 +41,7 @@ func main() {
         // ComputeExtendedDataSquare failed
     }
 
-    // Save all shares in flattended form.
-    // Note: slices returned from Row() an Column() are read-only.
+    // Save all shares in flattened form.
     // If you need to write to them, copy first.
     flattened := make([][]byte, 0, eds.Width()*eds.Width())
     for i := uint(0); i < eds.Width(); i++ {
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ func main() {
     fours := bytes.Repeat([]byte{4}, bufferSize)
 
     // Compute parity shares
-    eds, _ := rsmt2d.ComputeExtendedDataSquare(
+    eds, err := rsmt2d.ComputeExtendedDataSquare(
         [][]byte{
             ones, twos,
             threes, fours,
@@ -37,14 +37,16 @@ func main() {
         codec,
         rsmt2d.NewDefaultTree,
     )
+    if err != nil {
+        // ComputeExtendedDataSquare failed
+    }
 
     // Save all shares in flattended form.
     // Note: slices returned from Row() an Column() are read-only.
     // If you need to write to them, copy first.
     flattened := make([][]byte, 0, eds.Width()*eds.Width())
     for i := uint(0); i < eds.Width(); i++ {
-        startIndex := i * eds.Width()
-        copy(flattened[startIndex:startIndex+eds.Width()], eds.Row(i))
+        flattened = append(flattened, eds.Row(i)...)
     }
 
     // Delete some shares, just enough so that repairing is possible.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -56,7 +57,7 @@ func main() {
     // Repair square.
     repaired, err := rsmt2d.RepairExtendedDataSquare(
         eds.RowRoots(),
-        eds.ColumnRoots(),
+        eds.ColRoots(),
```
liamsi, line 60: I actually liked the written-out `ColumnRoots` better. Especially for public APIs (for local vars, and internal methods col is just fine). But I guess that is a matter of taste and I can get used to Col. I like that you changed it consistently in all places (as far as I can see). I feel consistency is more important here.

extendeddatacrossword.go
```diff
@@ -25,14 +25,14 @@ func (e *ErrByzantineRow) Error() string {
...
+type ErrByzantineCol struct {
+	ColNumber uint     // Column index
+	Shares    [][]byte // Pre-repaired column shares. Missing shares are nil.
 }
 
-func (e *ErrByzantineColumn) Error() string {
-	return fmt.Sprintf("byzantine column: %d", e.ColumnNumber)
+func (e *ErrByzantineCol) Error() string {
+	return fmt.Sprintf("byzantine column: %d", e.ColNumber)
 }
```
liamsi, line 36: I feel like we really should cover all these places in tests, too. But orthogonal to this PR.

extendeddatasquare.go
```diff
@@ -120,3 +128,34 @@ func (eds *ExtendedDataSquare) deepCopy(codec Codec) (ExtendedDataSquare, error)
 	eds, err := ImportExtendedDataSquare(eds.flattened(), codec, eds.createTreeFn)
 	return *eds, err
 }
+
+// Col returns a column slice.
+// This slice is a copy of the internal column slice.
+func (eds *ExtendedDataSquare) Col(y uint) [][]byte {
+	s := make([][]byte, eds.width)
+	copy(s, eds.colSlice(0, y, eds.width))
+	return s
```
liamsi, line 137: nit: append might be more readable/idiomatic? But again, mostly a style question and the `copy` is very explicit.

extendeddatasquare.go
```diff
@@ -120,3 +128,34 @@ func (eds *ExtendedDataSquare) deepCopy(codec Codec) (ExtendedDataSquare, error)
...
+}
```
liamsi, line 138: This new public method is untested?

rsmt2d_test.go
```diff
@@ -0,0 +1,61 @@
+package rsmt2d_test
```
liamsi, line 1: Nice! I think we should move most of the tests into this package. Can happen in a followup PR.

We should open an issue about testing (test-coverage and this) if it doesn't exist yet.

liamsi (approved):
Great! I think we can merge this as is. Though we should open an issue to make sure we test _all_ public APIs (including methods introduced here) and about test coverage.

extendeddatacrossword.go
adlerjohn: Agree, opened #58 to track increasing test coverage.

extendeddatasquare.go
adlerjohn: Tracking in #58.

rsmt2d_test.go
adlerjohn: Opened #58.

Issue #65, opened 2022-03-25
EDS is not modified in-place


adlerjohn:
Comment on repair says that EDS is modified in-place, but it's not. Fix comment.

PR #66, opened 2022-03-25
Return repaired eds in case of errors : ErrUnrepairableDataSquare,ErrByzantineRow, ErrByzantineCol

vgonkivs:
Related to #65 

As EDS is not modified in place, then we need to return it in case of errors.

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -15,6 +15,9 @@ const (
 // ErrUnrepairableDataSquare is thrown when there is insufficient chunks to repair the square.
 var ErrUnrepairableDataSquare = errors.New("failed to solve data square")
 
+// ErrNoChunksAvailable is thrown when dataSquare is empty
+var ErrNoChunksAvailable = errors.New("no chunks available")
+
 // ErrByzantineRow is thrown when a repaired row does not match the expected row Merkle root.
 type ErrByzantineRow struct {
 	RowNumber uint     // Row index
@@ -77,7 +80,7 @@ func RepairExtendedDataSquare(
 	}
 
 	if chunkSize == 0 {
-		return nil, ErrUnrepairableDataSquare
+		return nil, ErrNoChunksAvailable
 	}
 
 	fillerChunk := bytes.Repeat([]byte{0}, chunkSize)
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -25,6 +25,13 @@ func (e *ErrByzantineRow) Error() string {
 	return fmt.Sprintf("byzantine row: %d", e.RowNumber)
 }
 
+// checks if error equal to ErrByzantineRow/ErrByzantineCol
+func isByzantineError(err error) bool {
+	var errRow *ErrByzantineRow
+	var errCol *ErrByzantineCol
+	return errors.As(err, &errRow) || errors.As(err, &errCol)
+}
+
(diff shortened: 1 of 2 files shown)


extendeddatacrossword_test.go
```diff
@@ -56,10 +56,15 @@ func TestRepairExtendedDataSquare(t *testing.T) {
 		flattened[4], flattened[5], flattened[6], flattened[7] = nil, nil, nil, nil
 		flattened[8], flattened[9], flattened[10] = nil, nil, nil
 		flattened[12], flattened[13], flattened[14] = nil, nil, nil
-		_, err = RepairExtendedDataSquare(original.getRowRoots(), original.getColRoots(), flattened, codec, NewDefaultTree)
+
+		var extHeader *ExtendedDataSquare
+		extHeader, err = RepairExtendedDataSquare(original.getRowRoots(), original.getColRoots(), flattened, codec, NewDefaultTree)
```
liamsi, line 61: nit: this is not really a header, right? maybe just call the var repairedEDS, or, simply eds?

liamsi (approved):
Changes look good. One nit in the test-code.

extendeddatacrossword_test.go
vgonkivs: Renamed to `eds`

Wondertan (approved):
👍🏻

liamsi (approved):
👏🏼

extendeddatacrossword.go
```diff
@@ -88,15 +95,13 @@ func RepairExtendedDataSquare(
...
 		return nil, err
 	}
 
-	err = eds.solveCrossword(rowRoots, colRoots, bitMat, codec)
-	if err != nil {
-		return nil, err
-	}
-
-	return eds, err
+	return eds, eds.solveCrossword(rowRoots, colRoots, bitMat, codec)
```
adlerjohn, line 104: Can you add a test to show that the `eds` returned is actually the correct one here (i.e. the one _after_ `solveCrossword`)?

vgonkivs: done

extendeddatacrossword_test.go
```diff
@@ -121,23 +133,68 @@ func TestRepairExtendedDataSquare(t *testing.T) {
...
+		if err != nil {
+			t.Errorf("did not return a repaired EDS")
+		}
+		_ = edsNew.solveCrossword(corrupted.getRowRoots(), corrupted.getColRoots(), bitMat, codec)
+
+		if !reflect.DeepEqual(edsNew.ColRoots(), eds.ColRoots()) {
+			t.Fatal("col roots of EDSes do not match: eds after solving: ", edsNew.ColRoots(), ",eds after repairing: ", eds.ColRoots())
+		}
+		if !reflect.DeepEqual(edsNew.RowRoots(), eds.RowRoots()) {
+			t.Fatal("col roots of EDSes do not match: eds after solving: ", edsNew.RowRoots(), ", eds after repairing: ", eds.RowRoots())
```
adlerjohn, line 196:
```suggestion
			t.Fatal("row roots of EDSes do not match: eds after solving: ", edsNew.RowRoots(), ", eds after repairing: ", eds.RowRoots())
```

liamsi (approved):

adlerjohn (approved):

PR #70, opened 2022-04-12
Test all codecs in simple roundtrip test

liamsi:


rsmt2d_test.go
```diff
@@ -7,54 +10,65 @@ import (
...
-		rsmt2d.NewDefaultTree,
-	)
-	if err != nil {
-		t.Errorf("ComputeExtendedDataSquare failed")
+	tests := []struct {
+		name string
+		// Size of each share, in bytes
+		shareSize int
+		codec     rsmt2d.Codec
+	}{
```
musalbas, line 20: can you not just range over `codecs`?

liamsi: I could, yes. But now you can change the shareSize here too. Or you can add a different tree with one line etc. Table driven tests are considered best practice in golang.
https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

liamsi: It's just easier to extend that test. All tests should be like this ideally.

liamsi: https://github.com/golang/go/wiki/TableDrivenTests

adlerjohn (approved):

rsmt2d_test.go
liamsi: Also, in case of a test-failure you would get a better error message that tells you which sub-test fails. If you iterate over the codecs only, it would give you just the line where the assert failed without telling which codec is to blame (I just ran into this issue in the light of #72).

Issue #74, opened 2022-04-15
Remove Infectious in favor of Leopard only


adlerjohn:

liamsi:
Proposal: I think it would simplify the build flow, if we move the go-leopard wrapper into the rsmt2d library. 
It could still be its own module (with its own go.mod) but that way we can reference the build via LD_LIBRARY_PATH relative to the rsmt2d source root without having to install it in a more regular path (currently go-leopard aims for https://github.com/celestiaorg/go-leopard/blob/b7091fe12379ca80b59ed54b3fe9e03046a694d8/Makefile#L13).

liamsi:
In case we want to keep a go-library in here that does not require cgo, we should really move to @klauspost's. Infectious is essentially a very old version of his library augmented with forward error correction (see this [blog post](https://innovation.vivint.com/introduction-to-reed-solomon-bc264d0794f8), or simply the license and [credit](https://github.com/vivint/infectious#thanks) of the infectious lib. 
I quickly added klauspost's here and benchmarks look very promising as well: https://github.com/celestiaorg/rsmt2d/compare/klauspost?expand=1

liamsi:
Instead of removing infectious, we could create sub-packages, and making the user responsible to import/register the particular codec they want.

as @Wondertan said:
> The rsmt2d could have several subpkgs with Codec implementations, which are not registered automatically on the registry, but the user of rsmt2d could register what it needs(app). This way, old codecs would still live in the rsmt2d repo, with the app knowing nothing about them.

This could mean the user calling some "register" method. Or even just importing a certain sub-package.

rootulp:
Oops I created a duplicate issue https://github.com/celestiaorg/rsmt2d/issues/151

Proposal to close this issue because it discusses go-leopard which this repo no longer uses.

Issue #73, opened 2022-04-15
support maintained gf8 RS lib alternative


liamsi:
I think it would be worthwhile to use [this library](https://github.com/klauspost/reedsolomon) instead of [infectious](https://github.com/vivint/infectious) as a default GF8 lib. It is maintained and seems heavily optimized.

musalbas:
It's probably not as optimized as the leopard library though, which using a faster O(log(n)) FFT-based algorithm, whereas most Reed-Solomon libraries use a standard O(n^2) algorithm. We can't mix and match these libraries, because the leopard one uses a "Novel Polynomial Basis", which gives different encoding output.

liamsi:
Sure, I'm absolutely in favour of only having leopard in here if this is the only we want to use. Let's nuke infectious from the main branch entirely then.

PR #75, opened 2022-04-16
remove infectious, keep leopard only

liamsi:
ref: #73 
closes: #74


liamsi:
Converted back to draft because:
 - we can probably get rid of the leopard build tag if that is the default now anyways
 - need to ensure that the readme contains all info how to build this on all common architectures first

liamsi:
I think this can be closed as we are integrating klauspost's leopard implementation instead.

musalbas:
But why would switching to klauspost mean we shouldn't remove infectious?

Issue #72, opened 2022-04-15
Continuous repairing over the same square


Wondertan:
### What
The current state of the lib does not take into account a use case where the user executes `RepairExtendedSqaure` multiple times over the same data square slice. In such case, rsmt2d [fills with zeroes](https://github.com/celestiaorg/rsmt2d/blob/master/extendeddatacrossword.go#L86-L92) all the nill shares/chunks that are hashed on the subsequent repair preventing [sanity check](https://github.com/celestiaorg/rsmt2d/blob/9c00b79df4d4c58cbae7f14309728ce8dde0ab80/extendeddatacrossword.go#L359) to pass.

### Why
This is useful for repairing retries as new data comes from the network without doing additional copies of the square.

### How
As per @adlerjohn, the original intent to fill nil shares/chunks with zeroes was to match the library interface of the underlying codec implementations. However, we have codec wrappers/facades for each codec implementation that should independently adapt to the library interface without forcing the higher-level abstracted rsmt2d logic to match low-level APIs. 

#### TODO
- [ ] Change `Codec` interface contract to always accept nil shares instead of zeroed shares
- [ ] Investigate codec implementation libraries and their API codec; Find the difference and define which are different from the contract above
- [ ] Update local `Codec` implementation facades to match the contract above by adapting it to the underlying libraries API.

liamsi:
Decision:
 - use nils in RepairExtendedDataSquare and move the usage of zeroes to the specific codec impl when needed
 - this essentially makes the BitMask implementation and its usage obsolete

PR #76, opened 2022-04-16
Represent missing shares as nil internally

### LINKED ISSUE #71: Revert #66
#66 was based on the wrong assumption that the flattened data square slice given to `RepairExtendedDataSquare` _is not modified in place while it is_. In fact, the given slice is modified in place, and library users can and should access the repaired shares through it, so the change to make the `eds` accessible on `ErrByzantineRow/Col` is not needed anymore.

adlerjohn:
Fixes #71 #72

1. Bump Go version to 1.17
2. Fixes #71, reverting the changes made in #66 ([579abb7](https://github.com/celestiaorg/rsmt2d/pull/76/commits/579abb71c03178ac384944f5c49310185699eee1))
3. Remove bit matrix and instead simply keep `nil` shares as `nil`.
4. General cleanup and modernization (e.g. using range-based `for`s).
5. **Important**: fixed a bug where orthogonal cols/rows were never actually checked for correctness ([323f8fa](https://github.com/celestiaorg/rsmt2d/pull/76/commits/323f8fae6ad0a0208979a9f3d06b9ea27c564593)). Prior to this change, the condition to perform the check was essentially "if a share is missing and the col/row is complete" which is obviously never true; a col/row with a missing share is never complete. Fixing this exposed that the _rebuilt_ row/col was being used rather than the orthogonal col/row.


liamsi (commented):
Did a first pass and this looks very good to me! 🙌🏼  Before I approve I'll do a more thorough 2nd pass (soon).

extendeddatacrossword.go
```diff
@@ -415,10 +359,39 @@ func noMissingData(input [][]byte) bool {
 	return true
 }
 
+func numOnes(input [][]byte) int {
```
liamsi, line 362: Is this function used anywhere? If not, is there a reason to add it? Looks like we should delete it.

extendeddatacrossword.go
```diff
@@ -415,10 +359,39 @@ func noMissingData(input [][]byte) bool {
...
+	var counter int
+	for _, d := range input {
+		if d != nil {
+			counter++
+		}
+	}
+
+	return counter
+}
+
```
liamsi, line 372:
```suggestion
```

extendeddatacrossword.go
adlerjohn: Ah shoot I ported the bitmatrix functions over than forgot to remove that now-unused now 😂

infectiousRSGF8.go
```diff
@@ -19,10 +19,11 @@ func NewRSGF8Codec() *rsGF8Codec {
 	return &rsGF8Codec{make(map[int]*infectious.FEC)}
 }
 
-// Encode uses uses the infectous RSGF8 codec to encode the provided data
 func (c *rsGF8Codec) Encode(data [][]byte) ([][]byte, error) {
```
evan-forbes, line 22: why are we removing this comment?

rsmt2d_test.go
```diff
@@ -68,7 +67,84 @@ func TestEdsRepairRoundtripSimple(t *testing.T) {
...
+			err = rsmt2d.RepairExtendedDataSquare(
+				eds.RowRoots(),
+				eds.ColRoots(),
+				flattened,
+				tt.codec,
+				rsmt2d.NewDefaultTree,
+			)
+			if err != nil {
+				// Should now passing, since sufficient data.
+				t.Errorf("RepairExtendedDataSquare failed")
```
liamsi, line 145:
```suggestion
				t.Errorf("RepairExtendedDataSquare failed: %v", err)
```

rsmt2d_test.go
```diff
@@ -68,7 +67,84 @@ func TestEdsRepairRoundtripSimple(t *testing.T) {
...
+			err = rsmt2d.RepairExtendedDataSquare(
+				eds.RowRoots(),
+				eds.ColRoots(),
+				flattened,
+				tt.codec,
+				rsmt2d.NewDefaultTree,
+			)
+			if err == nil {
+				// Should fail since insufficient data.
+				t.Errorf("RepairExtendedDataSquare did not fail")
```
liamsi, line 131: nit: It would be even nicer if the test would communicate which error you are expecting here (if feasible).

liamsi (approved):
Awesome work 👏🏼

evan-forbes (approved):
wow, this is a great improvement! only one minor nit

infectiousRSGF8.go
adlerjohn: I added doctrings to the interface rather than the implementation.

rsmt2d_test.go
adlerjohn: Fixed in [6f587bc](https://github.com/celestiaorg/rsmt2d/pull/76/commits/6f587bc54aedbd39e7a2255707f2f6dcbd9e43a6)

rsmt2d_test.go
```diff
@@ -66,9 +65,86 @@ func TestEdsRepairRoundtripSimple(t *testing.T) {
...
+			err = rsmt2d.RepairExtendedDataSquare(
+				eds.RowRoots(),
+				eds.ColRoots(),
+				flattened,
+				tt.codec,
+				rsmt2d.NewDefaultTree,
+			)
+			if err == nil {
+				// Should fail since insufficient data.
+				t.Errorf("RepairExtendedDataSquare did not fail. Expected ErrUnrepairableDataSquare but got nil.")
```
liamsi, line 131: Optional: Better use the error directly instead of hard coding the string here. Also, we could assert that via errors.Is or errors.As.

liamsi: Or simply:
```suggestion
				t.Errorf("RepairExtendedDataSquare did not fail. Expected %v but got nil.", ErrUnrepairableDataSquare)
```

adlerjohn: Fixed in [5ffe833](https://github.com/celestiaorg/rsmt2d/pull/76/commits/5ffe83307d9ca2ca8af297160e42de27d89d28b9)

liamsi (approved):
(Still) excellent work :)

PR #79, opened 2022-04-25
Make repair a method on EDS

### LINKED ISSUE #77: api: rework `RepairExtendedSquare` to be `Repair` method on `ExtendedDataSquare`
This way, lib users can avoid reimporting the square repeatedly for cases where Repair has to be called multiple times.

adlerjohn:
Fixes #77

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -52,40 +52,30 @@ func (e *ErrByzantineCol) Error() string {
 // prior to the Byzantine row or column being repaired, and the Byzantine row
 // or column prior to repair is returned in the error with missing shares as
 // nil.
-func RepairExtendedDataSquare(
-	rowRoots [][]byte,
-	colRoots [][]byte,
-	data [][]byte,
+func (eds *ExtendedDataSquare) RepairExtendedDataSquare(
 	codec Codec,
 	treeCreatorFn TreeConstructorFn,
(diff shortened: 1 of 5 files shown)


extendeddatacrossword.go
```diff
@@ -52,29 +49,13 @@ func (e *ErrByzantineCol) Error() string {
 // prior to the Byzantine row or column being repaired, and the Byzantine row
 // or column prior to repair is returned in the error with missing shares as
 // nil.
-func RepairExtendedDataSquare(
+func (eds *ExtendedDataSquare) RepairExtendedDataSquare(
```
Wondertan, line 52: Let's name it just ‘Repair’. The struct itself already mentions that it is an extended data sqaure

vgonkivs:
LGTM. Agree with @Wondertan that `Repair` is enough.

vgonkivs (approved):

extendeddatacrossword.go
adlerjohn: Fixed in [54b8499](https://github.com/celestiaorg/rsmt2d/pull/79/commits/54b8499ca6ab761069e30a7b79b4025035984db4)

Wondertan (approved):

liamsi (approved):
👌🏼

PR #81, opened 2022-04-26
Add String method for Axis.

adlerjohn:
At @Wondertan's request.

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -36,14 +36,7 @@ type ErrByzantineData struct {
 }
 
 func (e *ErrByzantineData) Error() string {
-	switch e.Axis {
-	case Row:
-		return fmt.Sprintf("byzantine row: %d", e.Index)
-	case Col:
-		return fmt.Sprintf("byzantine column: %d", e.Index)
-	default:
-		panic(fmt.Sprintf("invalid axis: %d", e.Axis))
-	}
+	return fmt.Sprintf("byzantine %s: %d", e.Axis, e.Index)
 }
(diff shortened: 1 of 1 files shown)


Wondertan:
You can now remove the switch we introduced in last PR @adlerjohn

adlerjohn:
Don't we still need this `switch` to be able to identify which axis the error is for? The `String()` method should just be for printing out.

https://github.com/celestiaorg/rsmt2d/blob/35efa74fe884a5cbcdd90a8c410b5ffaeaa82770/extendeddatacrossword.go#L39-L46

adlerjohn:
Why remove this `switch`? It's for printing out the error.

https://github.com/celestiaorg/rsmt2d/blob/35efa74fe884a5cbcdd90a8c410b5ffaeaa82770/extendeddatacrossword.go#L39-L46

liamsi (approved):
I don't like that this and the prev. PR slightly decreased test-coverage but this lgtm but this is not blocking merging this PR. 

I also don't understand why switch in Axis.String() replaces the switch-case in ErrByzantineData.Error() as per Hlib's comment.

Wondertan:
@adlerjohn, @liamsi. The switch can be removed with:
 `return fmt.Sprintf("byzantine %s: %d", e.Axis, e.Index) `

adlerjohn:
Ah now I get it. Fixed in [609541b](https://github.com/celestiaorg/rsmt2d/pull/81/commits/609541b4cd7da14a539a8d6c27efb47170121d74)

PR #86, opened 2022-04-27
add extra build tags

evan-forbes:
without these build tags, the go compiler will sometimes complain, such as when the e2e test for tendermint is being compiled.

after merging, we will need to create a v0.5.1 release so that core can properly import this change.

### DIFF
--- a/leopard.go
+++ b/leopard.go
@@ -1,4 +1,5 @@
 //go:build leopard
+// +build leopard
 
 // Note that if the build tag leopard is used, liblibleopard.a
 // has to be present where the linker will find it.
--- a/rsmt2d_test.go
+++ b/rsmt2d_test.go
@@ -1,4 +1,5 @@
 //go:build leopard
+// +build leopard
 
 package rsmt2d_test
 


evan-forbes:
to make things more confusing, this PR seems to work fine

https://github.com/celestiaorg/celestia-core/pull/736

but when I try to use it in either of the tendermint upgrading PRs, it breaks :shrug:

adlerjohn:
Those are only needed for Go <1.17, which is unsupported in the `go.mod` and CI. There's no guarantee this library will work with Go <1.17, so there's not much point in adding those lines. Relevant: https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

Can the Tendermint CI tasks have their Go updated to 1.17 or above? I know @liamsi already updated some. Maybe your branch is outdated @evan-forbes?

evan-forbes:
ahh the docker container was using 1.16. Dang, assumed that tendermint 35 already updated to 1.17

thanks @adlerjohn!

liamsi:
Maybe we should have simply kept both type of flags until all deps and docker images updated to go 1.17.

PR #88, opened 2022-05-08
perf: check nil chunks directly to avoid copying

Wondertan:
`getCell` seems like another method that was supposed to become public but never got to it. I think so because it creates a copy of a chunk for encapsulation reasons, but It is only used internally now and in tests. This PR solves the problem that it avoids the use of `getCell` internally, thus avoiding copy internally, specifically in `...rangeMissingNoData`, which is just a helper func that should not do a copy for every chunk,

`getCell` won't be used anywhere after this PR is merged, so we should consider either removing it or publicizing it.

### DIFF
--- a/datasquare.go
+++ b/datasquare.go
@@ -231,7 +231,7 @@ func (ds *dataSquare) getColRoot(y uint) []byte {
 	return tree.Root()
 }
 
-// getCell returns a copy a single chunk at a specific cell.
+// getCell returns a copy of single chunk at a specific cell.
 func (ds *dataSquare) getCell(x uint, y uint) []byte {
 	if ds.squareRow[x][y] == nil {
 		return nil
--- a/datasquare.go
+++ b/datasquare.go
@@ -231,7 +231,7 @@ func (ds *dataSquare) getColRoot(y uint) []byte {
 	return tree.Root()
 }
 
-// getCell returns a single chunk at a specific cell.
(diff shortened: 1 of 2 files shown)


datasquare.go
```diff
@@ -231,7 +231,7 @@ func (ds *dataSquare) getColRoot(y uint) []byte {
 	return tree.Root()
 }
 
-// getCell returns a single chunk at a specific cell.
+// getCell returns a copy a single chunk at a specific cell.
```
adlerjohn, line 234:
```suggestion
// getCell returns a copy of single chunk at a specific cell.
```

adlerjohn (commented):
I would be in favor of exporting `GetCell` since there's not really any other clean way of getting the value of a cell without knowing internals.

adlerjohn:
@Wondertan are you going to export `GetCells` in this PR, or leave it to a subsequent issue?

liamsi (approved):
Change makes sense. Please follow up on @adlerjohn's comments of making getCell public.

Wondertan:
@adlerjohn, both work for me, but I think it is better to make GetCell in a separate PR, lemme make an issue for this

Wondertan:
Done https://github.com/celestiaorg/rsmt2d/issues/89

adlerjohn (approved):

Issue #85, opened 2022-04-27
api: Consider dropping EDS importing in favour of EDS constructor


Wondertan:
Currently, EDS can only be imported from a flattened data square slice and is updated in place, breaking encapsulation and allowing users to incorrectly change the underlying slice causing undefined behavior for the library if used incorrectly. Instead, we should provide a new constructor that allocates/manages the EDS independently and provide proper setters/getters(such as #83) to make the rsmt2d foolproof by hiding complexity and further allocation/performance optimizations behind the API.

Issue #84, opened 2022-04-27
api: cache imported roots for EDS


Wondertan:
Currently, the EDS has a stateless design, meaning that each operation(repairing) with it recomputes and verifies the integrity of the data according to the given roots. Instead, we can cache the roots during import and stick to those during the whole lifespan of an EDS instance. This is especially useful for users who try to repair the same EDS multiple times over the same roots.

Proposed Changes:
* Remove roots params from `Repair` and cache them on EDS
* Don't reset roots in `setCell` together with #83. Resetting was initially introduced to prevent users from breaking eds when using public `setCell`(which was actually never made). However, #83 proposes another solution for proper encapsulation(setting only nil cell) which makes resetting obsolete.

PR #92, opened 2022-06-11
Export `SetCell` and `GetCell`.

### LINKED ISSUE #83: api: Public `SetCell`
Currently, only a private `setCell` method is used internally to set repaired cells, while we can enable it publicly so that API users can fill cells/shares on the fly during the repairing process. Currently, a user has to alter the flattened data square slice imported with importing helper, which requires a relatively [complicated formula](https://github.com/celestiaorg/celestia-node/blob/main/ipld/retriever_quadrant.go#L83-L98)) to calculate cell position. Instead, users would benefit from public coordinates based `SetCell`.

Additionally, `SetCell` should allow setting values only for the nil cells to prevent breaking the integrity of the imported square verified by the imported and [cached roots.](https://github.com/celestiaorg/rsmt2d/issues/82). This also means that public `SetCell` won't need to reset roots like its private bro `setCell` [does](https://github.com/celestiaorg/rsmt2d/blob/master/datasquare.go#L247).

### LINKED ISSUE #89: api: publicalize `getCell`
Currently, `getCell` is private, but it is written with encapsulation in mind making it safe to be exposed, which we should do to provide users direct access to the shares.

adlerjohn:
Fixes #83
Fixes #89

Unlike the suggestion in #83 to not reset cached roots, that functionality was kept. The reason is that setting a cell from `nil` to non-`nil` does not guarantee that the roots match up with the leaves. The roots actually need to be re-computed to guarantee this, so the cache is cleared every time a cell is set.

### DIFF
--- a/datasquare.go
+++ b/datasquare.go
@@ -2,6 +2,7 @@ package rsmt2d
 
 import (
 	"errors"
+	"fmt"
 	"math"
 )
 
@@ -241,8 +242,19 @@ func (ds *dataSquare) GetCell(x uint, y uint) []byte {
 	return cell
 }
 
-// SetCell sets a specific cell.
+// SetCell sets a specific cell. Cell to set must be `nil`.
+// Panics if attempting to set a cell that is not `nil`.
 func (ds *dataSquare) SetCell(x uint, y uint, newChunk []byte) {
+	if ds.squareRow[x][y] != nil {
+		panic(fmt.Sprintf("cannot set cell (%d, %d) as it already has a value %x", x, y, ds.squareRow[x][y]))
+	}
+	ds.squareRow[x][y] = newChunk
+	ds.squareCol[y][x] = newChunk
+	ds.resetRoots()
+}
+
+// setCell sets a specific cell.
+func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) {
 	ds.squareRow[x][y] = newChunk
 	ds.squareCol[y][x] = newChunk
 	ds.resetRoots()
--- a/datasquare.go
+++ b/datasquare.go
@@ -117,7 +117,7 @@ func (ds *dataSquare) rowSlice(x uint, y uint, length uint) [][]byte {
 }
 
 // row returns a row slice.
-// Do not modify this slice directly, instead use setCell.
+// Do not modify this slice directly, instead use SetCell.
 func (ds *dataSquare) row(x uint) [][]byte {
 	return ds.rowSlice(x, 0, ds.width)
 }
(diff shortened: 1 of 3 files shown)


liamsi (approved):
Change looks good but did make we sure that with making the two of these public, we did not make the API less fool-proof (or easier to shoot yourself in the foot)?

Wondertan:
Why is this breaking @adlerjohn. It just adds new features to the API without breaking anything existing.

Wondertan:
> Unlike the suggestion in https://github.com/celestiaorg/rsmt2d/issues/83 to not reset cached roots, that functionality was kept. The reason is that setting a cell from nil to non-nil does not guarantee that the roots match up with the leaves. The roots actually need to be re-computed to guarantee this, so the cache is cleared every time a cell is set.

This omits the https://github.com/celestiaorg/rsmt2d/issues/84 from the picture. The #84 could be explained as "set expected roots once and ensure reconstruction matches them" :
* The cached roots are not the ones we compute inside the square but those we get from the square setup by the user. (I also mixed up those two while describing issues). 
	* We have roots from fixed DAH, and we just want to ensure the square we reconstruct matches it. 
	* We don't have new roots for each new Repair attempt, while the current API asks you to pass those each attempt. Or is there a use case where one square is repairable by different roots?
	* Maybe this also only makes sense with the custom constructor(#85) only.
* I agree resetting _computed_ roots is right even if we allow setting only non-nil shares. It does not make sense to reset the _expected_ roots, which I mixed up with the cached/computed.
* It still makes sense, though, to disallow setting(fool-prooving) non-nil cells to prevent computed, verified, or already set by the user cells from changing (within this PR).

adlerjohn:
> did make we sure that with making the two of these public, we did not make the API less fool-proof (or easier to shoot yourself in the foot)?

I don't see why it would be less foolproof. The alternative is setting cells manually then re-importing the data square, which is way more prone to screwing up.

> Why is this breaking

Ah you're correct, the public API is only additive.

> I agree resetting computed roots is right even if we allow setting only non-nil shares.

Yep, the resetting done here is for the computed roots.

> It still makes sense, though, to disallow setting(fool-prooving) non-nil cells to prevent computed, verified, or already set by the user cells from changing (within this PR).

That would make tests more cumbersome without an additional method to allow setting a cell to `nil`, but that can just be added!

adlerjohn:
@Wondertan fixed in [3910c4f](https://github.com/celestiaorg/rsmt2d/pull/92/commits/3910c4fb71d6dfca8ef408da96daafd2a9bc6a61)

Wondertan (approved):
Nice. Thanks.

PR #93, opened 2022-06-14
Cache codec and tree func.

### LINKED ISSUE #82: api: Cache codec and tree constructor in EDS after import
Importing of EDS requires a user to pass codec and tree constructor. Also, the `Repair` method requires passing these repeatedly on each repair retry, also allowing the user to change roots and codec on already imported eds. Instead, importing should cache those values and reuse them on `Repair`.

adlerjohn:
Fixes #82

Interestingly, the tree func was already cached, and the passed in tree func in `Repair` was unused.


Wondertan (approved):
Thanks. LGTM

PR #96, opened 2022-06-22
Hierarchical tests in extendeddatacrossword_test and datasquare_test

### LINKED ISSUE #51: Table-driven tests
Ref https://github.com/lazyledger/rsmt2d/pull/50#discussion_r615941444

distractedm1nd:
Adds subtests to `extendeddatacrossword_test.go` and `datasquare_test.go` - would this already be enough to close #51 ?

### DIFF
--- a/datasquare_test.go
+++ b/datasquare_test.go
@@ -9,36 +9,43 @@ import (
 )
 
 func TestNewDataSquare(t *testing.T) {
-	t.Run("1x1", func(t *testing.T) {
-		result, err := newDataSquare([][]byte{{1, 2}}, NewDefaultTree)
-		if err != nil {
-			panic(err)
-		}
-		if !reflect.DeepEqual(result.squareRow, [][][]byte{{{1, 2}}}) {
-			t.Errorf("newDataSquare failed for 1x1 square")
-		}
-	})
-	t.Run("2x2", func(t *testing.T) {
-		result, err := newDataSquare([][]byte{{1, 2}, {3, 4}, {5, 6}, {7, 8}}, NewDefaultTree)
-		if err != nil {
-			panic(err)
-		}
-		if !reflect.DeepEqual(result.squareRow, [][][]byte{{{1, 2}, {3, 4}}, {{5, 6}, {7, 8}}}) {
-			t.Errorf("newDataSquare failed for 2x2 square")
-		}
-	})
-	t.Run("InconsistentNumberOfChunks", func(t *testing.T) {
-		_, err := newDataSquare([][]byte{{1, 2}, {3, 4}, {5, 6}}, NewDefaultTree)
-		if err == nil {
-			t.Errorf("newDataSquare failed; inconsistent number of chunks accepted")
-		}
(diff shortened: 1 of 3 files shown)


extendeddatacrossword_test.go
```diff
@@ -20,139 +20,150 @@ type PseudoFraudProof struct {
...
+					}
+					corrupted.setCell(0, 0, corruptChunk)
+					err = corrupted.Repair(corrupted.getRowRoots(), corrupted.getColRoots())
+					if !errors.As(err, &byzData) || byzData.Axis != Row {
+						t.Errorf("did not return a ErrByzantineData for a bad row; got: %v", err)
+					}
+					// Construct the fraud proof
+					fraudProof := PseudoFraudProof{0, byzData.Index, byzData.Shares}
+					// Verify the fraud proof
+					// TODO in a real fraud proof, also verify Merkle proof for each non-nil share.
```
distractedm1nd, line 108: Should an issue be created out of this TODO?

adlerjohn: Hmmm. It's not really a TODO as much as it is a note. Constructing the Merkle proofs would only be possible for a subset of non-`nil` shares, and verifying them would be a lot of extra work that isn't directly related to rsmt2d.

datasquare_test.go
```diff
@@ -7,54 +7,62 @@ import (
...
+		if err == nil {
+			t.Errorf("newDataSquare failed; inconsistent number of chunks accepted")
+		}
+	})
+	t.Run("UnequalChunkSize", func(t *testing.T) {
+		_, err := newDataSquare([][]byte{{1, 2}, {3, 4}, {5, 6}, {7}}, NewDefaultTree)
+		if err == nil {
+			t.Errorf("newDataSquare failed; chunks of unequal size accepted")
+		}
+	})
```
evan-forbes, line 39: my interpretation of table driven tests would be to use a format similar to https://github.com/celestiaorg/rsmt2d/blob/d771ce746cea1dad73a475a4b2d51a311476b874/rsmt2d_test.go#L15-L27

I think this would allow us to avoid a lot of code repetition and hopefully increase readability. If all of the sub-tests in a given test cannot fit in that format, then perhaps those sub tests belong in their own test.

evan-forbes (commented):
left one comment that I think could be applied to most of the modifications here

thanks for digging into this @distractedm1nd !

datasquare_test.go
distractedm1nd: Yeah that makes a lot more sense - thanks! Will see what I can do there

adlerjohn (approved):
Looks fine to me, will defer to @liamsi for final review.

distractedm1nd:
Just needs another review

adlerjohn (commented):
merge conflict

liamsi (approved):
Merge conflict needs to be resolved. Otherwise, LGTM!

liamsi (approved):
👏🏼

PR #97, opened 2022-06-22
Fuzz Testing

### LINKED ISSUE #68: Add fuzzing tests
See for example: https://github.com/tendermint/tendermint/tree/master/test/fuzz

distractedm1nd:
Opening this as a draft PR, because I only added one fuzz test so we can get an idea of what it would look like before writing more. Closes #68.

I also exported `ds.Flattened`, since it is a useful data structure to have (for example as input to `ImportExtendedDataSquare`)

The two tests that were previously in this package are now implemented as subtests. When not running with `-fuzz`, they run with the original inputs provided in the `f.Add` call.

I removed the second half of the`TestEdsRepairTwice` test - since I assume the purpose is to test if the correct error is thrown when there are too many erasures. The case it tests for after repair is covered in the `MaximumErasures` subtest anyways. Or am I missing something?

### DIFF
--- a/datasquare.go
+++ b/datasquare.go
@@ -260,7 +260,8 @@ func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) {
 	ds.resetRoots()
 }
 
-func (ds *dataSquare) flattened() [][]byte {
+// Flattened returns the concatenated rows of the data square.
+func (ds *dataSquare) Flattened() [][]byte {
 	flattened := [][]byte(nil)
 	for _, data := range ds.squareRow {
 		flattened = append(flattened, data...)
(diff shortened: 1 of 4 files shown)


distractedm1nd:
@adlerjohn Yeah was debating that myself as well - makes sense, will do.

Btw, I believe the CI is failing because native go fuzzing is only supported since 1.18

evan-forbes:
> yeah, we'll have to update everything to use go 1.18 if we want to use the new fuzzer. we should do this in other repos as well, but it should be done in a separate PR imo

adlerjohn:
Agree, unless and until we want all downstream dependencies to require Go 1.18, we should hold off on this. Unless as you said, you can use a Go 1.18 module in a Go 1.17 project without issue.

liamsi:
Can we close this one until we continue working on it?

liamsi:
Closing as discussed

PR #100, opened 2022-07-05
api: element-wise copy on eds.Row and eds.Col

### LINKED ISSUE #99: api: `eds.Row` and `eds.Col` do not actually return copies
Since we work with nested slices, copy just copies the references to the inner slices. So a user modifying an `eds.Row` will actually modify the internal representation - this is not the intended behavior, correct? 

https://github.com/celestiaorg/rsmt2d/blob/e3daf260bc77f69ec3a4b7bbd4d1d7b13a6015fd/extendeddatasquare.go#L133-L152

distractedm1nd:
Closes #99


Wondertan (approved):
LGTM

Wondertan:
> Should I add tests for all the exported functions that have specific restraints on mutability? Can open a new PR for that

:yes:

PR #103, opened 2022-07-06
Switch Reed-Solomon implementation to github.com/klauspost/reedsolomon

elias-orijtech:
This is a work in progress until github/klauspost/reedsolomon is fast and stable enough to replace go-leopard. See https://github.com/klauspost/reedsolomon/pull/191 for the leopard port.

@liamsi what's the next step? Obviously performance optimizations to bring the Go port closer to the C++ version, but how much? I see references to the 8-bit version of leopard, do you need that ported as well or is klauspost/reedsolomon enough for total share counts <= 256?

CC @odeke-em .


liamsi:
>  I see references to the 8-bit version of leopard, do you need that ported as well [...]

No, the 16-bit version of leopard should be sufficient IMO. 

Will review this PR and https://github.com/klauspost/reedsolomon/pull/191 shortly. Thanks a lot! Really excited about this!

leopard.go
```diff
@@ -1,11 +1,8 @@
-//go:build leopard
```
adlerjohn, line 1: If we're removing this, then should the CI also be updated to not build with the `leopard` tag, and not require installing `libleopard`?

elias-orijtech: Done.

odeke-em:
@elias-orijtech thanks for working on this, nice work! @klauspost merged your changes per https://github.com/klauspost/reedsolomon/pull/191 so all good to go!

evan-forbes (approved):
I think this is ready to be merged :slightly_smiling_face: 

any other changes can be handled in a different PR imo

musalbas:
According to these benchmarks by @evan-forbes, this new implementation is 40x slower than go-leopard in rsmt2d. Could there be an issue with the way it's integrated in rsmt2d?

```
BenchmarkRepair/Repairing_128x128_ODS_using_RSGF8-24              	       1	16736594416 ns/op
BenchmarkExtension/RSGF8_size_128-24                               	      16	  66631976 ns/op

new pure go leopard
BenchmarkRepair/Repairing_128x128_ODS_using_LeopardFF16-24         	       1	25971504223 ns/op
BenchmarkExtension/LeopardFF16_size_128-24                         	       1	3378071341 ns/op

old c++ leopard
BenchmarkRepair/Repairing_128x128_ODS_using_LeopardFF16-24        	       1	16808911720 ns/op
BenchmarkExtension/LeopardFF16_size_128-24                        	      40	  28827064 ns/op```

**Comment by musalbas** [comment]
For encoding, according to the profiler, most of the time for encoding is spent on calling New(). Maybe the same re-encoder can be reused instead of creating a new one every time? It says here it can be reused: https://github.com/klauspost/reedsolomon/blob/master/reedsolomon.go#L387

profile001

File: leopard.go [hunk 2]
```diff
@@ -18,12 +15,37 @@ func init() {
 type leoRSFF8Codec struct{}
 
 func (l leoRSFF8Codec) Encode(data [][]byte) ([][]byte, error) {
-	return leopard.Encode(data)
+	return encode(data)
+}
+
+func encode(data [][]byte) ([][]byte, error) {
+	enc, err := reedsolomon.New(len(data), len(data))
```
liamsi, line 22: From https://github.com/celestiaorg/rsmt2d/pull/103#issuecomment-1243519416 it looks like we should be re-using the instance for same sizes. We need to be careful when using the instances in go-routines though.

klauspost:
Yes, encoders can/should be reused. Forward matrices are preserved between calls, and for gf8 inverse matrices as well. The encoder can safely be shared between goroutines.

You can [force GF16](https://github.com/klauspost/reedsolomon#forcing-leopard-gf16) for all sizes, but reconstruction is rather expensive.

musalbas:
Closing in favor of #111 which includes the commits here.

PR #111, opened 2022-09-12
Switch Reed-Solomon implementation to github.com/klauspost/reedsolomon, add cache encoder

rahulghangas:
Switches  `celestiaorg/go-leopard` with `klauspost/reedsolomon`

Implements a bounded, thread-safe double cache to store a new encoder. An encoder is kept in the front cache depending on how recently it was seen.

- [x] Tests for cache


musalbas:
This includes the commits for #103 so can't we close #103 now?

musalbas:
It should be noted that LeoRSFF8 isn't actually GF8 right now because there's no Leopard GF8 in klauspost right now, so that codec should be removed unless Leopard GF8 is implemented (unless we want to still include go-leopard).

utils/cache.go
```diff
@@ -0,0 +1,79 @@
...
+	"sync"
+)
+
+type Cache[T any] interface {
+	Insert(int, T)
+	Query(int) (T, bool)
+}
+
+var (
+	DefaultDoubleCacheCapacity = 1024 * 1024
```
evan-forbes, line 14: [question]
where does this value come from? 
[nit][not blocking]
mind adding a doc comment?

utils/cache.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	if value, ok := r.cacheBack[id]; ok {
+		r.cacheFrontSize += int(reflect.TypeOf(value).Size())
+		r.cacheFront[id] = value
+		return value, ok
+	}
+
+	var value T
+	return value, false
+}
```
evan-forbes, line 79: [not blocking][nit]
these are pretty simple, and they are already tested using the rest of the tests so I'm ok w/o adding additional unit tests, although some simple ones certainly wouldn't hurt for future reference. Especially since we're exporting these.

evan-forbes (commented):
left a few non-blocking comments. I'd approve after others get a chance to also take a look.

nice work!

utils/cache.go
rahulghangas: Just a random value. 1 MB of default cache?

utils/cache.go
rahulghangas: Done

evan-forbes (dismissed):
:rocket: :100:

musalbas:
We should merge this yet until this is resolved: https://github.com/celestiaorg/rsmt2d/pull/111#issuecomment-1245304954

evan-forbes:
@musalbas now that we have an issue to remind us to do this in the future #120, can we unblock this specific PR?

musalbas:
Not yet because as mentioned in #120, the current leopard defaults to GF8 when the shard count is <256, but the new leopard in this PR uses GF16 for shards <256. This results in different parity shares, and also the performance is not as high as it probably could be if it used GF8 for shard count <256.

evan-forbes:
> also the performance is not as high as it probably could be if it used GF8 for shard count <256.

evan-forbes:
> the current leopard defaults to GF8 when the shard count is <256, but the new leopard in this PR uses GF16 for shards <256. This results in different parity shares when the shard count is <256

Sorry if this is a stupid question, but why is having different parity shares an issue? aren't we also removing support for the old codec in this PR?

> ... and also the performance is not as high as it probably could be if it used GF8 for shard count <256.

can this be handled in a separate PR tho? I'm not seeing the benefit on increasing the scope of this PR.

musalbas:
>Sorry if this is a stupid question, but why is having different parity shares an issue? aren't we also removing support for the old codec in this PR?

It means that the LeoRSFF16 in this PR does not produce the same outputs as the LeoRSFF16 in the previous version. This is not a problem in itself as we're not at mainnet; however this would be mean that after mainnet we would not be able to change the codec for shards <256 to FF8, without a hard fork.

>can this be handled in a separate PR tho? I'm not seeing the benefit on increasing the scope of this PR.

IMO no because as mentioned above, this PR is a breaking change. Therefore we either need to make a conscious decision that this we will drop FF8 for the forseeable future and make a breaking change, or implement FF8 and make it not a breaking change. But IMO making one breaking change and then another breaking change to make it compatible again isn't a good idea.

evan-forbes:
I see, thanks for clarifying!

musalbas:
I think none of the caching functions here should be exported.

musalbas:
The klauspost reedsolomon library now supports GF8 as well as GF16. This should be enabled via `WithLeopardGF(true)` instead of `WithLeopardGF16(true)`: https://github.com/klauspost/reedsolomon/blob/24a9bd7f60b8c50d5ddda503370303ef1892c2c8/options.go#L254

leopard.go
```diff
@@ -18,12 +18,49 @@ func init() {
...
-	return leopard.Encode(data)
+	return encode(data)
+}
+
+func encode(data [][]byte) ([][]byte, error) {
+	dataLen := len(data)
+	enc, ok := encoderCache.Query(dataLen)
+	if !ok {
+		var err error
+		enc, err = reedsolomon.New(dataLen, dataLen, reedsolomon.WithLeopardGF16(true))
```
elias-orijtech, line 29: With https://github.com/klauspost/reedsolomon/pull/209 and @klauspost's optimization work, you can replace `WithLeopardGF16` with `WithLeopardGF` to get both GF16 and GF8 (depending on shard size).

musalbas:
see also: https://github.com/celestiaorg/rsmt2d/pull/75#issuecomment-1265622177

musalbas:
Do this caching functions need to be exported, if not could they be lowercased?

rahulghangas:
> Do this caching functions need to be exported, if not could they be lowercased? As this will make the API messy

Right, forgot about that. Will make that change

klauspost:
FYI, just released `v1.11.1`.

leopard.go
```diff
@@ -1,54 +1,63 @@
...
-func (l leoRSFF8Codec) Encode(data [][]byte) ([][]byte, error) {
-	return leopard.Encode(data)
+	registerCodec(Leopard, newLeoRSCodec())
 }
 
-func (l leoRSFF8Codec) Decode(data [][]byte) ([][]byte, error) {
-	half := len(data) / 2
-	return leopard.Decode(data[:half], data[half:])
+type leoRSCodec struct {
+	encoderCache *doubleCache[reedsolomon.Encoder]
```
Wondertan, line 14: @rahulghangas, have you used the [`sync.Pool`](https://pkg.go.dev/sync#Pool) primitive before? It seems like the purpose of the pool exactly matches the cache's goal. If not, there are various articles on the internet explaining the primitive. You can also check how @klauspost's reedsolomon library uses it for efficient reusage of allocated byte buffers.

Using Pool will give use:
* Less code to maintain and reinvent
* Non-blocking access to the pooled caches
	* The cache has a mutex which will harm performance for the parallelization of multiple EDS computes. The more, the worst it gets. 
* Does not use `reflect` pkg, which is slow in the way the cache uses it(`TypeOf`)

cache.go
```diff
@@ -0,0 +1,79 @@
...
+	opts.capacity = capacity / 2
+	return opts
+}
+
+type doubleCache[T any] struct {
+	opts           doubleCacheOptions
+	cacheMu        *sync.Mutex
+	cacheFrontSize int
+	cacheFront     map[int]T
+	cacheBack      map[int]T
```
liamsi, line 32: Maybe a dumb question but I do not fully understand why a double cache makes sense here. The access will be fast as internally we are accessing a map anyways (the complexity will asymptotically be the same in the front as well as in the back). Wouldn't single cache be as good as this and would be much simpler to follow? 
I almost feel a [sync.Map](https://pkg.go.dev/sync#Map) would have served the same purpose. Or maybe a tiny wrapper to make it bounded. But maybe I am missing something obvious here.

codecs.go
```diff
@@ -3,9 +3,8 @@ package rsmt2d
 import "fmt"
 
 const (
-	LeopardFF16 = "LeopardFF16"
-	LeopardFF8  = "LeopardFF8"
-	RSGF8       = "RSFG8"
+	Leopard = "Leopard"
+	RSGF8   = "RSFG8"
```
liamsi, line 7: Just noted when reviewing locally that this constant is not used properly. 
This is somewhat orthogonal to this PR but it should be used here IMO: https://github.com/celestiaorg/rsmt2d/blob/866085f9d14cbcf7dad8c3663c5d47d5ffa3621f/infectiousRSGF8.go#L12

codecs.go
```diff
@@ -29,16 +28,9 @@ func registerCodec(ct string, codec Codec) {
 	codecs[ct] = codec
 }
 
-func NewLeoRSFF16Codec() Codec {
-	if codec, has := codecs[LeopardFF16]; has {
+func NewLeoRSCodec() Codec {
+	if codec, has := codecs[Leopard]; has {
 		return codec
 	}
 	panic("cannot use codec LeopardFF16 without the 'leopard' build tag")
```
liamsi, line 35: This panic does not make any sense anymore!

cache.go
liamsi: I've used a sync.Map here: https://github.com/celestiaorg/rsmt2d/pull/136

Does the cache need to be strictly bounded? I think that is not really necessary as a rather small number of different sizes will occur over time and for Celestia specifically it will be a bunch of powers-of-two (cc @musalbas)

rahulghangas: The motivation behind the double cache was to keep it bounded and keep only encoders that are accessed frequently relevant

liamsi: 🤔  I think in practice less encoders than the default bound will be used anyways:
https://github.com/celestiaorg/rsmt2d/blob/e454f365b8e029fb147a656e2145dc916c1ba659/cache.go#L9

That said, I didn’t fully think this through but we should. Also what is the mem-footprint of one encoder.
Also I think the capacity isn’t really adjustable from the outside anymore. So this is more like a constant than an option in any case btw.

leopard.go
rahulghangas: Yes, ideally we should be using the same encoder per `dataLen`, and that is indeed the optimization here (insteda of creating a new encoder each time)

liamsi:
superseded by #136. Thanks a lot for the initial cache implementation @rahulghangas 👍🏼

PR #114, opened 2022-09-13
Crossword repair loop fixes

### LINKED ISSUE #109: Repairing a square is much slower than before
Due to a recent commit, repairing the 2D square is much slower than before (~18 seconds now vs ~0.3 seconds originally for a 128x128 original data square).

Now:
```
BenchmarkRepair/Repairing_128x128_ODS_using_LeopardFF8-16    	       1	18481421687 ns/op	3456429104 B/op	67648307 allocs/op
BenchmarkRepair/Repairing_128x128_ODS_using_LeopardFF16-16   	       1	18523533181 ns/op	3443092832 B/op	67388538 allocs/op
BenchmarkRepair/Repairing_128x128_ODS_using_RSGF8-16         	       1	19800162162 ns/op	3392873000 B/op	67115855 allocs/op
```

[Before](https://github.com/celestiaorg/rsmt2d/pull/35#discussion_r610870836):
```
BenchmarkRepair/Repairing_128x128_ODS_using_RSGF8-4         	       3	 379398435 ns/op
BenchmarkRepair/Repairing_128x128_ODS_using_LeopardFF8-4    	       4	 282697386 ns/op
BenchmarkRepair/Repairing_128x128_ODS_using_LeopardFF16-4   	       4	 293993599 ns/op
```

### LINKED ISSUE #113: solveCrosswordRow/Col isn't actually verifying that newly completed orthogonal vectors match their new merkle roots, as the eds is only updated after the check
https://github.com/celestiaorg/rsmt2d/blob/master/extendeddatacrossword.go#L203

musalbas:
- [x] Fix #113
- [x] Fix #109

Before:
```
goos: linux
goarch: amd64
pkg: github.com/celestiaorg/rsmt2d
cpu: 11th Gen Intel(R) Core(TM) i9-11900H @ 2.50GHz
BenchmarkRepair/Repairing_16x16_ODS_using_RSGF8-16         	      33	  36366217 ns/op
BenchmarkRepair/Repairing_32x32_ODS_using_RSGF8-16         	       4	 291625597 ns/op
BenchmarkRepair/Repairing_64x64_ODS_using_RSGF8-16         	       1	2196795856 ns/op
BenchmarkRepair/Repairing_128x128_ODS_using_RSGF8-16       	       1	18349927611 ns/op
```

After:
```
goos: linux
goarch: amd64
pkg: github.com/celestiaorg/rsmt2d
cpu: 11th Gen Intel(R) Core(TM) i9-11900H @ 2.50GHz
BenchmarkRepair/Repairing_16x16_ODS_using_RSGF8-16         	     363	   3733920 ns/op
BenchmarkRepair/Repairing_32x32_ODS_using_RSGF8-16         	      82	  20524141 ns/op
BenchmarkRepair/Repairing_64x64_ODS_using_RSGF8-16         	      16	  64611698 ns/op
BenchmarkRepair/Repairing_128x128_ODS_using_RSGF8-16       	       4	 268046116 ns/op
```

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -145,7 +145,11 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 	// Check that newly completed orthogonal vectors match their new merkle roots
 	for c := 0; c < int(eds.width); c++ {
 		col := eds.col(uint(c))
-		if noMissingData(col) {
+		if col[r] != nil {
+			continue // not newly completed
+		}
+		col[r] = rebuiltShares[c]
+		if noMissingData(col) { // not completed
 			err := eds.verifyAgainstColRoots(colRoots, uint(c), col)
 			if err != nil {
 				return false, false, err
@@ -203,7 +207,11 @@ func (eds *ExtendedDataSquare) solveCrosswordCol(
 	// Check that newly completed orthogonal vectors match their new merkle roots
 	for r := 0; r < int(eds.width); r++ {
 		row := eds.row(uint(r))
-		if noMissingData(row) {
+		if row[c] != nil {
+			continue // not newly completed
+		}
+		row[c] = rebuiltShares[r]
+		if noMissingData(row) { // completed
 			err := eds.verifyAgainstRowRoots(rowRoots, uint(r), row)
 			if err != nil {
 				return false, false, err
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -211,7 +211,7 @@ func (eds *ExtendedDataSquare) solveCrosswordCol(
 			continue // not newly completed
 		}
 		row[c] = rebuiltShares[r]
-		if noMissingData(row) { // completed
+		if noMissingData(row) { // not completed
 			err := eds.verifyAgainstRowRoots(rowRoots, uint(r), row)
 			if err != nil {
 				return false, false, err


extendeddatacrossword.go
```diff
@@ -145,7 +145,11 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 	// Check that newly completed orthogonal vectors match their new merkle roots
 	for c := 0; c < int(eds.width); c++ {
 		col := eds.col(uint(c))
-		if noMissingData(col) {
+		if col[r] != nil {
+			continue // not newly completed
+		}
+		col[r] = rebuiltShares[c]
```
evan-forbes, line 151: [question]
I think we're inserting some of the rebuilt shares twice, once here and then a second time on line 161. Should we instead insert the rebuilt shares before performing this step?

same goes for the change below

evan-forbes (commented):
Good find!! only had one question before I approve.

extendeddatacrossword.go
musalbas: I don't think so, because the EDS is only meant to contain the "last known" good square, so we can't insert the cell into the EDS until we know it's good.

evan-forbes: that makes sense!

evan-forbes (approved):

Wondertan:
@musalbas Can we write a test asserting that #113 is being fixed or make an issue that needs to be tested? It is indeed fixed, and a unit test to concrete this would not hurt.

musalbas:
Agreed

musalbas:
@adlerjohn can you also review this?

adlerjohn (commented):
Problem and fix make sense to me.

PR #115, opened 2022-09-13
Benchmark the root generation when benchmarking extension.

musalbas:
This results in more accurate extension benchmarking, computing the Merkle roots is a significant overhead.


evan-forbes (approved):
makes sense

PR #117, opened 2022-09-14
Fix #112

### LINKED ISSUE #112: ErrByzantineData includes bad shares
It seems that ErrByzantineData also includes bad shares _after_ reconstruction, when I believe the expected behavior (as expected by celestia-node) is that it should only return shares that are known to be good (eg Merkle proofs for them are known).

musalbas:
Fixes #112.

Note: we need to add substantially more tests for fraud proofs in rsmt2d as these cases aren't being tested.

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -139,6 +139,9 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 	// Check that rebuilt shares matches appropriate root
 	err = eds.verifyAgainstRowRoots(rowRoots, uint(r), rebuiltShares)
 	if err != nil {
+		if byzErr, ok := err.(*ErrByzantineData); ok {
+			byzErr.Shares = shares
+		}
 		return false, false, err
 	}
 
@@ -201,6 +204,9 @@ func (eds *ExtendedDataSquare) solveCrosswordCol(
 	// Check that rebuilt shares matches appropriate root
 	err = eds.verifyAgainstColRoots(colRoots, uint(c), rebuiltShares)
 	if err != nil {
+		if byzErr, ok := err.(*ErrByzantineData); ok {
+			byzErr.Shares = shares
+		}
 		return false, false, err
 	}
 
@@ -268,7 +274,7 @@ func (eds *ExtendedDataSquare) verifyAgainstRowRoots(
 	root := eds.computeSharesRoot(shares, r)
 
 	if !bytes.Equal(root, rowRoots[r]) {
-		return &ErrByzantineData{Row, r, shares}
+		return &ErrByzantineData{Row, r, nil}
 	}
 
 	return nil
@@ -282,7 +288,7 @@ func (eds *ExtendedDataSquare) verifyAgainstColRoots(
 	root := eds.computeSharesRoot(shares, c)
 
 	if !bytes.Equal(root, colRoots[c]) {
-		return &ErrByzantineData{Col, c, shares}
+		return &ErrByzantineData{Col, c, nil}
 	}
 
 	return nil
@@ -318,7 +324,7 @@ func (eds *ExtendedDataSquare) prerepairSanityCheck(
 				return err
 			}
 			if !bytes.Equal(flattenChunks(parityShares), flattenChunks(eds.rowSlice(i, eds.originalDataWidth, eds.originalDataWidth))) {
-				return &ErrByzantineData{Row, i, eds.row(i)}
+				return &ErrByzantineData{Row, i, eds.rowSlice(i, 0, eds.originalDataWidth)}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


musalbas:
Seems this fails with `-tags leopard`.

extendeddatacrossword.go
```diff
@@ -318,7 +324,7 @@ func (eds *ExtendedDataSquare) prerepairSanityCheck(
 				return err
 			}
 			if !bytes.Equal(flattenChunks(parityShares), flattenChunks(eds.rowSlice(i, eds.originalDataWidth, eds.originalDataWidth))) {
-				return &ErrByzantineData{Row, i, eds.row(i)}
+				return &ErrByzantineData{Row, i, eds.rowSlice(i, 0, eds.originalDataWidth)}
```
Wondertan, line 327: What if shares without Merkle proofs are part of the ODS?

musalbas:
I fixed the failing test, which was failing because when we only return the original shares in the sanity check where we have the full row/col, so I changed it back to returning the whole row/col.

It should be noted that the Shares in ErrByzantineData should be guaranteed to only be shares that we can generate Merkle proofs for; but they may be _more shares than necessary_ to prove fraud. Therefore currently it should be up to the application to strip out extraneous shares.

Wondertan:
>  so I changed it back to returning the whole row/col.

This makes ErrByzantineData ambiguous as now there are two different behaviors for the same error over the same API. Is there any other to fix the issue with the test?

musalbas:
It's not different behaviors, it's actually the same behaviour, and the fix actually makes the behaviour more consistent. The behaviour is that ErrByzantineData always returns *all* shares in a row or column that Merkle proofs can be generated for (i.e. they are in a fully completed orthogonal row or column).

Wondertan:
Ok, I misunderstood your original message. LGTM then

musalbas:
@adlerjohn please review

extendeddatacrossword.go
```diff
@@ -139,6 +139,9 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 	// Check that rebuilt shares matches appropriate root
 	err = eds.verifyAgainstRowRoots(rowRoots, uint(r), rebuiltShares)
 	if err != nil {
+		if byzErr, ok := err.(*ErrByzantineData); ok {
```
adlerjohn, line 142: You should either use `errors.Is` or `errors.As` according to golang: https://pkg.go.dev/errors

adlerjohn (commented):
While technically not necessary since the EDS only includes shares that are provable by the caller, it's not a bad change. Left one code style nit.

musalbas:
>While technically not necessary since the EDS only includes shares that are provable by the caller, it's not a bad change. Left one code style nit.

Actually, the EDS may include recovered shares that are not in the Merkle tree and thus are not provable by the caller, so this change is necessary. Do you agree? See https://github.com/celestiaorg/celestia-node/issues/1086

adlerjohn:
The _EDS_ shouldn't (if it does, then that's unintended), since recovered shares are only added to the EDS if they match the Merkle roots.

musalbas:
>since recovered shares are only added to the EDS if they match the Merkle roots

But the EDS itself adds recovered shares to itself (via the crossword), which might not match the Merkle root.

extendeddatacrossword.go
Wondertan: I also wanted to comment on this, but as long as it expects a specific type in here to amend it - it's fine. Though I think the way we set shares in it is rather hacky, but don't have anything else to propose as an easy fix

adlerjohn:
Inserting is only done after Merkle roots are matched.

https://github.com/celestiaorg/rsmt2d/blob/76b270f80f0b9ac966c6f6b043e31514574f90f3/extendeddatacrossword.go#L139-L145

https://github.com/celestiaorg/rsmt2d/blob/76b270f80f0b9ac966c6f6b043e31514574f90f3/extendeddatacrossword.go#L160

musalbas:
Hence "the returned EDS does not return those recovered shares" in my original message

extendeddatacrossword.go
musalbas: Done

extendeddatacrossword.go
```diff
@@ -138,8 +138,9 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 
 	// Check that rebuilt shares matches appropriate root
 	err = eds.verifyAgainstRowRoots(rowRoots, uint(r), rebuiltShares)
+	var byzErr *ErrByzantineData
 	if err != nil {
-		if byzErr, ok := err.(*ErrByzantineData); ok {
+		if errors.As(err, &byzErr) {
```
Wondertan, line 143: Move inside if case?

musalbas: wdym?

Wondertan: I mean byzErr should be declared after error check and not before

musalbas: Done

Wondertan (approved):

PR #126, opened 2022-09-21
chore!: Revert tree.Push api to only use []byte and not also the axis

### LINKED ISSUE #125: Remove the square index from the `tree.Push` method
Now that we're passing the axis to the constructor of the tree, we no longer have to pass the axis to the `tree.Push([]byte, SquareIndex)` method. This means we can revert to the normal interface for pushing data to a tree `tree.Push([]byte)`, as the implementation can keep track of the share index being pushed internally.

evan-forbes:
closes #125

### DIFF
--- a/datasquare.go
+++ b/datasquare.go
@@ -222,8 +222,8 @@ func (ds *dataSquare) getRowRoot(x uint) []byte {
 	}
 
 	tree := ds.createTreeFn(Row, x)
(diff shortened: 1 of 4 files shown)


Wondertan (approved):
👍🏻

adlerjohn (approved):

Issue #123, opened 2022-09-15
Investigate and potentially fix the multi-threading issue with decoding using the klauspost leopard implementation.\


evan-forbes:
When parallelising the crossword loop, it works fine with infectious, but errors on go-leopard and klauspost leopard. Seems like a potential issue with doing parallel decodings. Otherwise, it resulted in 4x+ performance increase for 128x128 blocks with infectious.

Branch with parallel decoding (klauspost): https://github.com/celestiaorg/rsmt2d/blob/parallelisation_klauspost/extendeddatacrossword.go#L73

klauspost leopard:
```
goos: linux
goarch: amd64
pkg: github.com/celestiaorg/rsmt2d
cpu: 11th Gen Intel(R) Core(TM) i9-11900H @ 2.50GHz
BenchmarkRepair/RSGF8_4x4x256_ODS-16         	--- FAIL: BenchmarkRepair/RSGF8_4x4x256_ODS-16
    extendeddatacrossword_test.go:217: byzantine row: 3
BenchmarkRepair/LeopardFF8_4x4x256_ODS-16    	     518	   2273441 ns/op
BenchmarkRepair/LeopardFF16_4x4x256_ODS-16   	     499	   2318601 ns/op
BenchmarkRepair/RSGF8_8x8x256_ODS-16         	--- FAIL: BenchmarkRepair/RSGF8_8x8x256_ODS-16
    extendeddatacrossword_test.go:217: byzantine row: 7
BenchmarkRepair/LeopardFF8_8x8x256_ODS-16    	     300	   3929318 ns/op
BenchmarkRepair/LeopardFF16_8x8x256_ODS-16   	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5781d4]

goroutine 436131 [running]:
github.com/klauspost/reedsolomon.mulgf16_avx2({0xc0092c6d00, 0x100, 0x100}, {0x0, 0x100, 0x100}, 0xc0048ccc80)
	/home/mus/go/pkg/mod/github.com/klauspost/reedsolomon@v1.11.0/galois_gen_amd64.s:63559 +0x54
github.com/klauspost/reedsolomon.mulgf16({0xc0092c6d00?, 0x10?, 0x10?}, {0x0?, 0x100?, 0xc000280000?}, 0xffff?, 0x0?)
[...]

PR #128, opened 2022-09-28
fix: check for nil before setting nil

Wondertan:
This change is motivated by the race conditions I observed. The roots are reset every time we do `SetCell` by setting nil. However, it causes race conditions between multiple `SetCell` calls. This change fixes this and is tested against celestia-node

### DIFF
--- a/datasquare.go
+++ b/datasquare.go
@@ -176,8 +176,12 @@ func (ds *dataSquare) setColSlice(x uint, y uint, newCol [][]byte) error {
 }
 
 func (ds *dataSquare) resetRoots() {
-	ds.rowRoots = nil
-	ds.colRoots = nil
+	if ds.rowRoots != nil {
+		ds.rowRoots = nil
+	}
+	if ds.colRoots != nil {
+		ds.colRoots = nil
+	}
 }
 
 func (ds *dataSquare) computeRoots() {


adlerjohn (commented):
Also confused at how this resolves race conditions. It's non-obvious (if at all), so should be documented in the code. Otherwise someone will go in and optimize those lines away without context, since there's no test to exercise the behavior in this repo.

Wondertan:
Happy to document this in the code.

Wondertan:
Done

rahulghangas:
Ran into the same issue while working on #123, can confirm it fixes the race condition

datasquare.go
```diff
@@ -176,8 +176,15 @@ func (ds *dataSquare) setColSlice(x uint, y uint, newCol [][]byte) error {
 }
 
 func (ds *dataSquare) resetRoots() {
-	ds.rowRoots = nil
-	ds.colRoots = nil
+	// don't write nil if it's already nil
+	// this prevents rewriting nil into shared memory slot
+	// when resetRoots is used from multiple routines
+	if ds.rowRoots != nil {
```
adlerjohn, line 182: Instead of this, could we use a mutex? E.g. functions other than `SetCell` use a mutex prior to calling `resetRoots`

https://github.com/celestiaorg/rsmt2d/blob/ed5b81ebc9a0262ea08834a50ef05501bc8b74fc/datasquare.go#L165-L166

Wondertan: The mutex is blocking. In my case there are hundreds of routines resetting the roots by calling SetCell. Contention between those should be avoided. The data structure in the nose around the EDs is non-blocking, and this solution keeps it so.

Orthogonally to multithreading, I feel like rewriting nil all the time is not a good thing to either, but I couldn't find solid proof for that.

Wondertan: Right now the EDS does not have any guarantees regarding thread safety for reads and writes. If we introduce it, we should do it not just in one place but more holistically. The parallelization is internal for the API and orthogonal. 

Additionally, we can extract the SyncEDS from the node, so the lib has a synchronized yet nonblocking version of the square.

adlerjohn (approved):
utACK, but would like @liamsi to confirm

Wondertan:
@liamsi, kind ping

liamsi (approved):
https://github.com/celestiaorg/rsmt2d/pull/128/files#r988034202

> Right now the EDS does not have any guarantees regarding thread safety for reads and writes.

Can you please open an issue around that?

Wondertan:
@liamsi, https://github.com/celestiaorg/rsmt2d/issues/135

PR #131, opened 2022-09-28
fix: parallel repair

### LINKED ISSUE #123: Investigate and potentially fix the multi-threading issue with decoding using the klauspost leopard implementation.\

rahulghangas:
`solveCrosswordRow` sets cells for the crossword which are accessed in `solveCrosswordColumn`, which means that both being run in parallel has a race condition.

This PR resolves this by running all `solveCrosswordRow` in parallel and waiting for all these calls to finish before executing all `solveCrosswordColumn` calls in parallel

- [x] Fixes #123

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -84,7 +84,7 @@ func (eds *ExtendedDataSquare) solveCrossword(
 		// Track if a single iteration of this loop made progress
 		progressMade := false
 
-		var trackerMutex sync.Mutex
+		var statusMutex sync.Mutex
 		// Loop through every row and column, attempt to rebuild each row or column if incomplete
 		for i := 0; i < int(eds.width); i++ {
 			i := i
@@ -95,8 +95,8 @@ func (eds *ExtendedDataSquare) solveCrossword(
 					return err
 				}
 
-				trackerMutex.Lock()
-				defer trackerMutex.Unlock()
+				statusMutex.Lock()
+				defer statusMutex.Unlock()
 				solved = solved && solvedRow
 				progressMade = progressMade || progressMadeRow
 				return nil
@@ -116,8 +116,8 @@ func (eds *ExtendedDataSquare) solveCrossword(
 					return err
 				}
 
-				trackerMutex.Lock()
-				defer trackerMutex.Unlock()
+				statusMutex.Lock()
+				defer statusMutex.Unlock()
 				solved = solved && solvedCol
 				progressMade = progressMade || progressMadeCol
 				return nil
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -84,7 +84,7 @@ func (eds *ExtendedDataSquare) solveCrossword(
 		// Track if a single iteration of this loop made progress
 		progressMade := false
 
-		var mut sync.Mutex
+		var trackerMutex sync.Mutex
 		// Loop through every row and column, attempt to rebuild each row or column if incomplete
 		for i := 0; i < int(eds.width); i++ {
 			i := i
@@ -95,8 +95,8 @@ func (eds *ExtendedDataSquare) solveCrossword(
 					return err
 				}
 
-				mut.Lock()
-				defer mut.Unlock()
+				trackerMutex.Lock()
+				defer trackerMutex.Unlock()
 				solved = solved && solvedRow
 				progressMade = progressMade || progressMadeRow
 				return nil
@@ -116,8 +116,8 @@ func (eds *ExtendedDataSquare) solveCrossword(
 					return err
 				}
 
-				mut.Lock()
-				defer mut.Unlock()
+				trackerMutex.Lock()
+				defer trackerMutex.Unlock()
 				solved = solved && solvedCol
 				progressMade = progressMade || progressMadeCol
 				return nil
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -85,15 +85,11 @@ func (eds *ExtendedDataSquare) solveCrossword(
 		progressMade := false
 
 		var mut sync.Mutex
-		var wg sync.WaitGroup
-		wg.Add(int(eds.width))
-
 		// Loop through every row and column, attempt to rebuild each row or column if incomplete
 		for i := 0; i < int(eds.width); i++ {
 			i := i
 
 			errs.Go(func() error {
-				defer wg.Done()
 				solvedRow, progressMadeRow, err := eds.solveCrosswordRow(i, rowRoots, colRoots)
 				if err != nil {
 					return err
@@ -106,8 +102,15 @@ func (eds *ExtendedDataSquare) solveCrossword(
 				return nil
 			})
 
+		}
+		if err := errs.Wait(); err != nil {
+			return err
+		}
+
+		for i := 0; i < int(eds.width); i++ {
+			i := i
+
 			errs.Go(func() error {
-				wg.Wait()
 				solvedCol, progressMadeCol, err := eds.solveCrosswordCol(i, rowRoots, colRoots)
 				if err != nil {
 					return err
@@ -120,7 +123,6 @@ func (eds *ExtendedDataSquare) solveCrossword(
 				return nil
 			})
 		}
-
 		if err := errs.Wait(); err != nil {
 			return err
 		}
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -186,6 +186,8 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 		if col[r] != nil {
 			continue // not newly completed
 		}
+
+		eds.rowColMutex[c].Lock()
 		col[r] = rebuiltShares[c]
 		if noMissingData(col) { // not completed
 			err := eds.verifyAgainstColRoots(colRoots, uint(c), col)
@@ -193,11 +195,14 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 				return false, false, err
 			}
 		}
+		eds.rowColMutex[c].Unlock()
 	}
 
 	// Insert rebuilt shares into square.
 	for c, s := range rebuiltShares {
+		eds.rowColMutex[c].Lock()
 		eds.setCell(uint(r), uint(c), s)
+		eds.rowColMutex[c].Unlock()
 	}
 
 	return true, true, nil
@@ -248,6 +253,8 @@ func (eds *ExtendedDataSquare) solveCrosswordCol(
 		if row[c] != nil {
 			continue // not newly completed
 		}
+
+		eds.rowColMutex[r].Lock()
 		row[c] = rebuiltShares[r]
 		if noMissingData(row) { // not completed
 			err := eds.verifyAgainstRowRoots(rowRoots, uint(r), row)
@@ -255,11 +262,14 @@ func (eds *ExtendedDataSquare) solveCrosswordCol(
 				return false, false, err
 			}
 		}
+		eds.rowColMutex[r].Unlock()
 	}
 
 	// Insert rebuilt shares into square.
 	for r, s := range rebuiltShares {
+		eds.rowColMutex[r].Lock()
 		eds.setCell(uint(r), uint(c), s)
+		eds.rowColMutex[r].Unlock()
 	}
 
 	return true, true, nil
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 
 	"golang.org/x/sync/errgroup"
 )
@@ -83,6 +84,8 @@ func (eds *ExtendedDataSquare) solveCrossword(
 		// Track if a single iteration of this loop made progress
 		progressMade := false
 
+		var mut sync.Mutex
+
 		// Loop through every row and column, attempt to rebuild each row or column if incomplete
 		for i := 0; i < int(eds.width); i++ {
 			i := i
@@ -93,6 +96,8 @@ func (eds *ExtendedDataSquare) solveCrossword(
 					return err
 				}
 
+				mut.Lock()
+				defer mut.Unlock()
 				solved = solved && solvedRow
 				progressMade = progressMade || progressMadeRow
 				return nil
@@ -104,6 +109,8 @@ func (eds *ExtendedDataSquare) solveCrossword(
 					return err
 				}
 
+				mut.Lock()
+				defer mut.Unlock()
 				solved = solved && solvedCol
 				progressMade = progressMade || progressMadeCol
 				return nil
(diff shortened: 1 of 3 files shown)


rahulghangas:
Linking conversation from #129 and #130

> This seems to imply the reason why parallelisation_crosswordloop didn't work was not because of the decoder, but because of rsmt2d. In that case how come it worked with infectious but not leopard or klauspost?

Infectious was failing on my machine, although sporadically. I think it was just the race condition at play. 

> Blocking column repair until row repair is complete means that we aren't maximizing the parallelisation that we can do. Is there another way to fix this, eg add mutexes for setting cells?

I don't think so, the race condition is between reading and writing of cells. For eg. `solveCrosswordRow` is setting cells [here](https://github.com/celestiaorg/rsmt2d/pull/130/files#diff-43b1ac1c622a799f2a850a10dee280ccfd5c04f315c69eb8848d215363dd8688L176) while the same cell is being read by `solveCrosswordCol` [here](https://github.com/celestiaorg/rsmt2d/pull/130/files#diff-43b1ac1c622a799f2a850a10dee280ccfd5c04f315c69eb8848d215363dd8688L212). iiuc, the cell must set with the rebuilt shares before being read

extendeddatacrossword.go
```diff
@@ -83,27 +84,37 @@ func (eds *ExtendedDataSquare) solveCrossword(
...
 
+				mut.Lock()
+				defer mut.Unlock()
 				solved = solved && solvedRow
 				progressMade = progressMade || progressMadeRow
 				return nil
 			})
 
 			errs.Go(func() error {
+				wg.Wait()
```
musalbas, line 110: Instead of adding wait groups and subroutines for inactive tasks would it be simpler if we just had two for loops, one for rows and one for columns?

musalbas:
>I don't think so, the race condition is between reading and writing of cells. For eg. solveCrosswordRow is setting cells [here](https://github.com/celestiaorg/rsmt2d/pull/130/files#diff-43b1ac1c622a799f2a850a10dee280ccfd5c04f315c69eb8848d215363dd8688L176) while the same cell is being read by solveCrosswordCol [here](https://github.com/celestiaorg/rsmt2d/pull/130/files#diff-43b1ac1c622a799f2a850a10dee280ccfd5c04f315c69eb8848d215363dd8688L212). iiuc, the cell must set with the rebuilt shares before being read

Makes sense - if the data is malicious then we might end up setting different data into the same cell

rahulghangas:
All data races have been fixed. #128 needs to be merged for tests to pass

@musalbas @adlerjohn please review

musalbas:
Can you explain why row/col mutexes has been added? From the original description of the race condition this doesn't seem necessary

rahulghangas:
> Can you explain why row/col mutexes has been added? From the original description of the race condition this doesn't seem necessary

After setting a cell with the rebuilt share, the entire row/col must be read to verify if the entire row has been set or not, [ref](https://github.com/celestiaorg/rsmt2d/pull/131/files#diff-43b1ac1c622a799f2a850a10dee280ccfd5c04f315c69eb8848d215363dd8688L176-L182)

While the entire row is being read, another go routine might be writing to the same row (just a different cell), which might lead to spurious results

rahulghangas:
`solveCrosswordRow` hits all columns [here](https://github.com/celestiaorg/rsmt2d/pull/131/files#diff-43b1ac1c622a799f2a850a10dee280ccfd5c04f315c69eb8848d215363dd8688R191-R197) and `solveCrosswordCol` hits all rows [here](https://github.com/celestiaorg/rsmt2d/pull/131/files#diff-43b1ac1c622a799f2a850a10dee280ccfd5c04f315c69eb8848d215363dd8688R258-R264)

musalbas:
I see, I think I can see what the race condition is, but I think that might still not fix it. Suppose the last two rows to complete a column are being repaired, but they both call `noMissingData` before rebuilt shares are inserted into the square. Wouldn't `noMissingData` return false for both so the orthogonal row/col will never get verified? We should think about what would fix it fully - maybe applying a mutex over the entire second half of solveCrosswordRow/Col?

adlerjohn (commented):
~#128 has been merged, maybe pull from default branch?~ Edit: wrong one, ignore.

rahulghangas:
> Suppose the last two rows to complete a column are being repaired, but they both call noMissingData before rebuilt shares are inserted into the square. Wouldn't noMissingData return false for both so the orthogonal row/col will never get verified?

I don't think so, because [here](https://github.com/celestiaorg/rsmt2d/pull/131/files#diff-43b1ac1c622a799f2a850a10dee280ccfd5c04f315c69eb8848d215363dd8688L176) the `dataCol` of the eds gets mutated because [this](https://github.com/celestiaorg/rsmt2d/pull/131/files#diff-43b1ac1c622a799f2a850a10dee280ccfd5c04f315c69eb8848d215363dd8688L172) is returning a slice, not a copy. Since the mutex is on the column, whichever row completes the column will successfully call `noMissingData` on it

musalbas:
Oh that's a bug then, `col[r] = rebuiltShares[c]` is not supposed to mutate the the underlying DS. Assuming we fix that bug would the aforementioned issue exist and how can it be solved?

musalbas:
On another general note about the mutex - I think instead of adding rowColMutex to the EDS, we should define rowColMutex in solveCrossword and pass it as arguements to solveCrosswordRow/Col, because the mutex is only relevant to crossword solving.

Unless we think there's a way to fix this by making EDS itself entirely thread safe - but my current intuition is that even if EDS was thread safe the bug would still exist because it's a race condition in the crossword solver itself rather than the underlying EDS.

rahulghangas:
>On another general note about the mutex - I think instead of adding rowColMutex to the EDS, we should define rowColMutex in solveCrossword and pass it as arguements to solveCrosswordRow/Col, because the mutex is only relevant to crossword solving

Agreed, that would be better

>Unless we think there's a way to fix this by making EDS itself entirely thread safe - but my current intuition is that even if EDS was thread safe the bug would still exist because it's a race condition in the crossword solver itself rather than the underlying EDS

Yes, the issue exists regardless of whether EDS is thread safe or not

> maybe applying a mutex over the entire second half of solveCrosswordRow/Col
This seems like this would be the best way forward with the given implementation

musalbas (changes requested):
instead of adding rowColMutex to the EDS, we should define rowColMutex in solveCrossword and pass it as arguements to solveCrosswordRow/Col, because the mutex is only relevant to crossword solving

rahulghangas:
Blocked by #134

musalbas:
Is there a reason why it's a single mutex now instead of one per row/col? Computing the Merkle root in computeSharesRoot takes up a significant chunk of the time so it would be ideal to parallelize that.

rahulghangas:
> I see, I think I can see what the race condition is, but I think that might still not fix it. Suppose the last two rows to complete a column are being repaired, but they both call noMissingData before rebuilt shares are inserted into the square. Wouldn't noMissingData return false for both so the orthogonal row/col will never get verified? We should think about what would fix it fully - maybe applying a mutex over the entire second half of solveCrosswordRow/Col?

We can't separate the insertion logic from the validation logic, since we get the race condition mentioned above

rootulp:
Going to close this b/c it doesn't appear to be actively being worked on.

PR #134, opened 2022-10-04
fix: eds mutation

### LINKED ISSUE #132: solveCrosswordRow/Col mutates underlying data square, when it shouldn't be
See https://github.com/celestiaorg/rsmt2d/pull/131#issuecomment-1264669180

See also https://github.com/celestiaorg/rsmt2d/issues/118

rahulghangas:
Call to `eds.row` returns a slice to the underlying row/col data. Hence, when rebuilt shares are added to this slice, the underlying data is being mutated before correct validation is done. On the other hand, using `eds.Row` is too expensive, since it makes a copy of the underlying data. This PR fixes the first issue while avoiding expensive copying by passing the rebuilt share and/or index to relevant methods

- [x] Fixes #132

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -117,7 +117,7 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 	rowRoots [][]byte,
 	colRoots [][]byte,
 ) (bool, bool, error) {
-	isComplete := noMissingData(eds.row(uint(r)))
+	isComplete := noMissingData(eds.row(uint(r)), -1)
 	if isComplete {
 		return true, false, nil
 	}
@@ -140,7 +140,7 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 	}
 
 	// Check that rebuilt shares matches appropriate root
-	err = eds.verifyAgainstRowRoots(rowRoots, uint(r), rebuiltShares)
+	err = eds.verifyAgainstRowRoots(rowRoots, uint(r), rebuiltShares, -1, nil)
 	if err != nil {
 		var byzErr *ErrByzantineData
 		if errors.As(err, &byzErr) {
@@ -155,9 +155,8 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 		if col[r] != nil {
 			continue // not newly completed
 		}
-		col[r] = rebuiltShares[c]
-		if noMissingData(col) { // not completed
-			err := eds.verifyAgainstColRoots(colRoots, uint(c), col)
+		if noMissingData(col, r) { // not completed
+			err := eds.verifyAgainstColRoots(colRoots, uint(c), col, r, rebuiltShares[c])
 			if err != nil {
 				return false, false, err
 			}
@@ -182,7 +181,7 @@ func (eds *ExtendedDataSquare) solveCrosswordCol(
 	rowRoots [][]byte,
 	colRoots [][]byte,
 ) (bool, bool, error) {
-	isComplete := noMissingData(eds.col(uint(c)))
+	isComplete := noMissingData(eds.col(uint(c)), -1)
 	if isComplete {
 		return true, false, nil
 	}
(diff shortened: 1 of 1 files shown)


musalbas:
Is this still ready for review @rahulghangas? I recall you said you had an idea about something yesterday

rahulghangas:
I thought about how to replace `-1` with something better, but didn't come up with anything. Thoughts?

Wondertan:
@musalbas, I liked your solution in https://github.com/celestiaorg/rsmt2d/pull/114 so much that removing it now while introducing more ugly branchiness in the code is disappointing.

liamsi:
> I thought about how to replace -1 with something better, but didn't come up with anything. Thoughts?


Maybe make `-1` a constant with a proper name s.t. where it used it is semantically clearer what it stands for?

rahulghangas:
@musalbas I think I can request only one review, so pinging you here. Do the new changes work?

musalbas:
If we can't think of a cleaner way it's ok with me. `NoShareInsertion` shouldn't be exported though.

extendeddatacrossword.go
```diff
@@ -15,6 +15,8 @@ type Axis int
 const (
 	Row Axis = iota
 	Col
+
+	NoShareInsertion = -1
```
liamsi, line 19: Agreed that this should not to be exported. Everything else LGTM.

rahulghangas: Resolved

liamsi (approved):
LGTM. One question though: would it be too difficult to test the change in bahviour as it only affects internals of the implementation?

rahulghangas:
The only behaviour that has changed (bee corrected) is for the state of the underlying eds when trying to repair from incorrect data. I guess we should add a test for that

liamsi:
> I guess we should add a test for that

Yes, please do 👍🏼

extendeddatacrossword.go
```diff
@@ -140,7 +142,7 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 	}
 
 	// Check that rebuilt shares matches appropriate root
-	err = eds.verifyAgainstRowRoots(rowRoots, uint(r), rebuiltShares)
+	err = eds.verifyAgainstRowRoots(rowRoots, uint(r), rebuiltShares, noShareInsertion, nil)
```
liamsi, line 145: Nit: I am not sure `verifyAgainstRowRoots` needs both `rebuiltIndex == -1` and `rebuiltShare == nil` to indicate computing the shares root without a previously rebuilt share.

extendeddatacrossword.go
```diff
@@ -274,9 +274,16 @@ func (eds *ExtendedDataSquare) rebuildShares(
...
 	rowRoots [][]byte,
 	r uint,
-	shares [][]byte,
+	oldShares [][]byte,
+	rebuiltIndex int,
+	rebuiltShare []byte,
 ) error {
-	root := eds.computeSharesRoot(shares, Row, r)
+	var root []byte
+	if rebuiltIndex < 0 || rebuiltShare == nil {
```
liamsi, line 282: Question: Isn't `rebuiltShare == nil` sufficient here to indicate this branch? Or can it happen in other cases?

liamsi: Ah, nvm.

adlerjohn (changes requested):

evan-forbes:
is this still being worked on? if so what is left?

extendeddatacrossword.go
```diff
@@ -274,9 +274,16 @@ func (eds *ExtendedDataSquare) rebuildShares(
 func (eds *ExtendedDataSquare) verifyAgainstRowRoots(
 	rowRoots [][]byte,
 	r uint,
-	shares [][]byte,
+	oldShares [][]byte,
+	rebuiltIndex int,
```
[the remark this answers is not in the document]
rahulghangas: Separating out leads to a lot of code duplication because `noShareInsertion` needs to be propagated all the way inside to the last function call. Which is why I included `noShareInsertion` in the first place

adlerjohn (approved):

PR #136, opened 2022-10-11
Switch to klauspost/reedsolomon implementation and cache encoder

liamsi:
Superseds #111 

Diff to #111 are these two commits:
 - https://github.com/celestiaorg/rsmt2d/pull/136/commits/d4a6f0929ee9ad7ecd77861c93d8dd59e73e15a0
 - https://github.com/celestiaorg/rsmt2d/pull/136/commits/a24c38ad5376f79ba345715e0c2bad1f3f0569de


musalbas:
Is this ready for review, if so can it be marked as such as it current in "draft".

musalbas:
What is the justification of using this instead of the double cache btw?

liamsi:
> Is this ready for review, if so can it be marked as such as it currently in "draft" and there are TODOs in the code.

I've marked this as draft because I am not 100% sure this is the better approach as compared to #111. I added you and others as reviewers for input on this. Particularly, I'm asking @rahulghangas for feedback. I want to push this over the finish line this or next week ideally. If we agree that the double cache is not really a hard requirement, then I'm very much in favour of this approach.

There is only one TODO added, namely:
https://github.com/celestiaorg/rsmt2d/blob/a24c38ad5376f79ba345715e0c2bad1f3f0569de/leopard.go#L16-L24
This can only be resolved much later. 

> What is the justification of using this instead of the double cache btw?

Less code, less code to maintain, much easier to understand. Not sure the double-cache is really necessary. See:
https://github.com/celestiaorg/rsmt2d/pull/111#discussion_r992261630

rahulghangas:
@liamsi given that the number of different data sizes are going to be relatively few, I am indeed in favor of this approach

codec_test.go
```diff
@@ -15,6 +15,7 @@ func BenchmarkEncoding(b *testing.B) {
 	// generate some fake data
 	data := generateRandData(128)
 	for codecName, codec := range codecs {
+		codec.Encode(data)
```
rahulghangas, line 18: I think I forgot to add a comment here to indicate that this is for making sure that the relevant encoder is cached

rahulghangas (approved):
LGTM!

codec_test.go
liamsi: Hmm, I wonder if that has any impact on the encoding speed for some codecs if it already ran on the same data 🤔 I'll add a comment and make sure that the init-run is with different data. Just in case.

liamsi: done in [9c35d77](https://github.com/celestiaorg/rsmt2d/pull/136/commits/9c35d7748872aeea319b4c4ba1ef880f404ce0c3)

leopard.go
```diff
@@ -1,54 +1,72 @@
...
+	//
+	// TODO: switch to a generic version of sync.Map with type reedsolomon.Encoder
+	// once it made it into the standard lib
+	encCache sync.Map
 }
 
-func newLeoRSFF8Codec() leoRSFF8Codec {
-	return leoRSFF8Codec{}
-}
+func (l leoRSCodec) Encode(data [][]byte) ([][]byte, error) {
```
Wondertan, line 27: blocking bug: we should use a pointer here! Otherwise, we copy `leoRSCodec` and the whole `sync.Map`, while it [must not be copied](https://pkg.go.dev/sync#Map).

Wondertan (commented):
There is a bug here. Look for the last comment

leopard.go
liamsi: Makes sense! Thanks for catching this 🙏🏼

liamsi: Fixed in [a424318](https://github.com/celestiaorg/rsmt2d/pull/136/commits/a424318e88877896bd1fa15e4b5d893c07f9a779)

leopard.go
```diff
@@ -1,54 +1,72 @@
...
+	dataLen := len(data)
+	enc, ok := l.encCache.Load(dataLen)
+	if !ok {
+		var err error
+		enc, err = reedsolomon.New(dataLen, dataLen, reedsolomon.WithLeopardGF(true))
+		if err != nil {
+			return nil, err
+		}
+		l.encCache.Store(dataLen, enc)
+	}
```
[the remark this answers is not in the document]
liamsi: resolved in https://github.com/celestiaorg/rsmt2d/pull/136/commits/02d187594de083cd53813e9dfc514fb506b56ffb

leopard.go
```diff
@@ -1,54 +1,72 @@
...
+	return shards[dataLen:], nil
 }
 
-func (leo leoRSFF16Codec) Decode(data [][]byte) ([][]byte, error) {
+func (l leoRSCodec) Decode(data [][]byte) ([][]byte, error) {
 	half := len(data) / 2
-	return leopard.Decode(data[:half], data[half:])
+	enc, ok := l.encCache.Load(half)
+	var err error
```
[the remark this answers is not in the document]
liamsi: also resolved in https://github.com/celestiaorg/rsmt2d/pull/136/commits/02d187594de083cd53813e9dfc514fb506b56ffb

liamsi:
Thanks for the thorough review @Wondertan!

leopard.go
rahulghangas: Good catch! I wasn't aware of this

Wondertan (approved):
LGTM

rsmt2d_test.go
```diff
@@ -86,8 +83,7 @@ func TestEdsRepairTwice(t *testing.T) {
 		shareSize int
 		codec     rsmt2d.Codec
 	}{
-		{"leopardFF8", bufferSize, rsmt2d.NewLeoRSFF8Codec()},
-		{"leopardFF16", bufferSize, rsmt2d.NewLeoRSFF16Codec()},
+		{"klauspost/leopard", bufferSize, rsmt2d.NewLeoRSCodec()},
```
musalbas, line 86: This is called `klauspost/leopard` here but `leopard` above?

liamsi: That is just the name of the test. But makes sense to make it consistent: https://github.com/celestiaorg/rsmt2d/pull/136/commits/1689a464eccea6c1f665f8e4612eac49640ecf2f

evan-forbes (approved):
:+1: :+1:

PR #140, opened 2023-01-03
feat: implementing `json.(Un)Marshaller` for `ExtendedDataSquare`

### LINKED ISSUE #139: Implement `json.(Un)Marshaller` for EDS
## Context

In celestia-node, we are introducing a new [`Getter` interface](https://github.com/celestiaorg/celestia-node/issues/1488). One of its methods `GetEDS` allows users to get full EDS over the network. Additionally, we want to expose this interface and the method over public API. Besides direct access over Go API, the Public API is also exposed over JSON-RPC. This approach allows any app dev to depend on native Go's interfaces and enjoy their ergonomics, while, in fact, talking to a remote server over JSON- RPC. However, it requires all the types in the method's signature to be JSON encodable. 

## Motivation

We at celestia-node want to give a top-notch DevEx to the ecosystem and believe that relying on well-defined canonical data structures, like EDS, in the API unlocks the best experience without limitations, so we don't want to make a wrapper EDS structure to implement JSON encoding, but instead, extend the existing definition of EDS.

## Open Questions

* Do we want to marshall the full EDS or only its quadrant?
	* A time-space tradeoff that we may allow power users to configure.
* Should we prevent marshaling when EDS integrity has not been verified? 
	* In practice, the celestia-node API will guarantee this, but we may want to additionally enforce this in the rsmt2d library

distractedm1nd:
## Overview
Closes #139

## Checklist

- [X] New and updated code has appropriate documentation
- [X] New and updated code has new and/or updated testing
- [X] Linked issues closed with keywords

### DIFF
--- a/extendeddatasquare_test.go
+++ b/extendeddatasquare_test.go
@@ -2,6 +2,7 @@ package rsmt2d
 
 import (
 	"crypto/rand"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"testing"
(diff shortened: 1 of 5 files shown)


extendeddatasquare_test.go
```diff
@@ -26,6 +27,31 @@ func TestComputeExtendedDataSquare(t *testing.T) {
...
 }
 
+func TestMarshalJSON(t *testing.T) {
+	codec := NewRSGF8Codec()
+	result, err := ComputeExtendedDataSquare([][]byte{
+		{1}, {2},
+		{3}, {4},
+	}, codec, NewDefaultTree)
+	if err != nil {
+		panic(err)
```
renaynay, line 37: y panic and not just failnow? you want stacktrace? ( in general in tests for rsmt2d )

renaynay (approved):
one q

extendeddatasquare_test.go
distractedm1nd: Probably for the stacktrace - this is how it's done in the other tests

Wondertan (approved):
Lgtm

liamsi (approved):

adlerjohn (approved):
utACK

Issue #156, opened 2023-03-22
Consider adding error return parameters for `Tree` interface methods


rootulp:
https://github.com/celestiaorg/rsmt2d/blob/33a57631a7fcdffaad54baa69b4371d9b2bac8df/tree.go#L20

Since this method has no error return type, the nmt_wrapper.go must `panic` on scenarios where it would make more sense to return an error. E.g. [here](https://github.com/celestiaorg/celestia-app/blob/ead76d2bb607ac8a2deaba552de86d4df74a116b/pkg/wrapper/nmt_wrapper.go#L94-L96)

rootulp:
This issue can be expanded to
https://github.com/celestiaorg/rsmt2d/blob/33a57631a7fcdffaad54baa69b4371d9b2bac8df/tree.go#L21

PR #157, opened 2023-03-31
chore!: add error return params to tree interface

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/156


datasquare.go
```diff
@@ -198,12 +198,20 @@ func (ds *dataSquare) computeRoots() {
...
+			}
+			rowRoots[i] = rowRoot
 		}(i)
 
 		go func(i uint) {
 			defer wg.Done()
-			colRoots[i] = ds.getColRoot(i)
+			colRoot, err := ds.getColRoot(i)
+			if err != nil {
+				panic(err)
```
evan-forbes, line 212: I know this probably shouldn't get hit and its in a goroutine, but is there a way to bubble these up as well? computing the roots is

evan-forbes (commented):
nice! thanks for doing this, I think its a really good change!

only had a single comment, but otherwise LGTM :+1:

datasquare.go
Wondertan: errgroup

rootulp: Thanks for the tip @Wondertan ! Resolved in https://github.com/celestiaorg/rsmt2d/pull/157/commits/bbacca88a5448fa16df25f9aeeece7e86c535fff

evan-forbes:
cc @Wondertan as this api change would have an effect on node

evan-forbes (approved):

datasquare.go
```diff
@@ -187,29 +189,41 @@ func (ds *dataSquare) resetRoots() {
 	}
 }
 
-func (ds *dataSquare) computeRoots() {
-	var wg sync.WaitGroup
+func (ds *dataSquare) computeRoots() error {
+	g := new(errgroup.Group)
```
Wondertan, line 193: New forces allocation on heap. There is no reason to do it here as it used within methods boundaries and should be kept on stack by var

rootulp: Does https://github.com/celestiaorg/rsmt2d/pull/157/commits/1956b1665b2669f945ecf23ce31d64d1211dd989 resolve your concern?

Wondertan: Yes

Wondertan: The quickest link I found: https://stackoverflow.com/a/10866871

Wondertan (approved):

evan-forbes (approved):

rootulp:
It looks like @Wondertan doesn't have write permissions on this repo so his approval doesn't satisfy the 2 approvals requirement. I removed @MSevey 's review request b/c he's on parental leave so this needs @liamsi 's review.

Wondertan:
And @liamsi is ooo as well, maybe @musalbas ?

rootulp:
Maybe @cmwaters has write permission to this repo b/c he's a member of core/app

SweeXordious (approved):

Issue #164, opened 2023-06-16
Unnecessary index calculation


andrijamitrovic23:
In the function `erasureExtendRow` when [setting a row slice](https://github.com/celestiaorg/rsmt2d/blob/6515446be623987ded119e530900589f30e4c69e/extendeddatasquare.go#L164), the start index calculation `len(shares)-int(eds.originalDataWidth)` is unnecessary because `len(shares)` is the same as the `eds.originalDataWidth`.

Same problem exists in the [erasureExtendCol](https://github.com/celestiaorg/rsmt2d/blob/6515446be623987ded119e530900589f30e4c69e/extendeddatasquare.go#LL167C1-L167C1) function as well.

rootulp:
> because `len(shares)` is the same as the `eds.originalDataWidth`

Appears correct b/c https://github.com/rootulp/rsmt2d/blob/2e59fbaa585e44d26cc294edb89b51af48dc61eb/codecs.go#L15-L17

PR #167, opened 2023-06-16
chore: remove unnecessary slice indexing

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/164

### DIFF
--- a/extendeddatasquare.go
+++ b/extendeddatasquare.go
@@ -154,28 +154,22 @@ func (eds *ExtendedDataSquare) erasureExtendSquare(codec Codec) error {
 }
 
 func (eds *ExtendedDataSquare) erasureExtendRow(codec Codec, i uint) error {
-	var shares [][]byte
-	var err error
-
-	shares, err = codec.Encode(eds.rowSlice(i, 0, eds.originalDataWidth))
+	parityShares, err := codec.Encode(eds.rowSlice(i, 0, eds.originalDataWidth))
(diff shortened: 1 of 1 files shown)


extendeddatasquare.go
```diff
@@ -154,25 +154,19 @@ func (eds *ExtendedDataSquare) erasureExtendSquare(codec Codec) error {
...
-	var shares [][]byte
-	var err error
-
-	shares, err = codec.Encode(eds.rowSlice(i, 0, eds.originalDataWidth))
+	parityShares, err := codec.Encode(eds.rowSlice(i, 0, eds.originalDataWidth))
 	if err != nil {
 		return err
 	}
-	return eds.setRowSlice(i, eds.originalDataWidth, shares[len(shares)-int(eds.originalDataWidth):])
+	return eds.setRowSlice(i, eds.originalDataWidth, parityShares)
```
evan-forbes, line 161: is parity shares the entire of row of shares, but we're only interested in the later half? or is this actually only the parity shares? not sure if it matters too much, but if it is the former, should we change the name to parity?

evan-forbes (approved):
quick question on var naming, but otherwise LGTM

extendeddatasquare.go
rootulp: `codec.Encode` returns just the parity shares so we don't need to index into it.

staheri14 (approved):

Issue #170, opened 2023-06-19
Check chunk size


andrijamitrovic23:
When setting cell in [SetCell](https://github.com/celestiaorg/rsmt2d/blob/6515446be623987ded119e530900589f30e4c69e/datasquare.go#L289) and [setCell](https://github.com/celestiaorg/rsmt2d/blob/6515446be623987ded119e530900589f30e4c69e/datasquare.go#L299-L303) a check that size of `newChunk` is equal to size of other chunks in data square should be done. If it is not equal, an error as in [newDataSquare](https://github.com/celestiaorg/rsmt2d/blob/6515446be623987ded119e530900589f30e4c69e/datasquare.go#LL35C1-L44C3) should be returned.

Issue #127, opened 2022-09-28
Issue with panic in `SetCell`


Wondertan:
## Context
Today, we [panic](https://github.com/celestiaorg/rsmt2d/blob/master/datasquare.go#L269-L271) when we attempt to set a non-nil cell. However, there is a case where this becomes problematic. The logic in celestia-node runs the Repair method several times, which partially repairs the EDS and only some rows and cols on each run. The networking logic does not know which ones were repaired, and may download a share that was already repaired, subsequently setting into the square and _panic_.

## Workaround
The workaround is to call `GetCell` before `SetCell` to check if the cell is already there. However, the function does [unnecessary copying](https://github.com/celestiaorg/rsmt2d/blob/master/datasquare.go#L262) inside.

## Possible solutions
* `HasCell` that only report true or false
  * This could work, but it's not clear why we need to panic and perform the check twice: (1) `HasCell` by user and in `SetCell` in the actual logic
```go
if !square.HasCell(x, y) {
  square.SetCell(x, y, cell)
}
```
* Don't panic in `SetCell` and return `bool` instead, which reports whether setting a cell succeeded

adlerjohn:
> and return `bool` instead

Should it return an `error` instead?

Wondertan:
@adlerjohn, It could. Don't think it really matters. From a high level, I would say it's fine to make it an acceptable case as we inherently have two commits to every cell. The networking layer, in the worst case, uses both to get and set the share, and it seems normal for rsmt2d to be aware of this by not erroring on such a case. Error and panic are something incorrect, while this case does seem to be expected.

Anyway, I don't feel strongly about this and am happy to go with any solution you feel is the best.

PR #172, opened 2023-06-19
chore!: check newChunk size

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/170
Closes https://github.com/celestiaorg/rsmt2d/issues/127

Breaking b/c modifies the function signature of a public function: `SetCell`

### DIFF
--- a/datasquare.go
+++ b/datasquare.go
@@ -284,22 +284,31 @@ func (ds *dataSquare) GetCell(x uint, y uint) []byte {
 	return cell
 }
 
-// SetCell sets a specific cell. Cell to set must be `nil`.
-// Panics if attempting to set a cell that is not `nil`.
-func (ds *dataSquare) SetCell(x uint, y uint, newChunk []byte) {
+// SetCell sets a specific cell. The cell to set must be `nil`. Returns an error
+// if the cell to set is not `nil` or newChunk is not the correct size.
+func (ds *dataSquare) SetCell(x uint, y uint, newChunk []byte) error {
 	if ds.squareRow[x][y] != nil {
-		panic(fmt.Sprintf("cannot set cell (%d, %d) as it already has a value %x", x, y, ds.squareRow[x][y]))
+		return fmt.Errorf("cannot set cell (%d, %d) as it already has a value %x", x, y, ds.squareRow[x][y])
+	}
+	if len(newChunk) != int(ds.chunkSize) {
+		return fmt.Errorf("cannot set cell with chunk size %d because dataSquare chunk size is %d", len(newChunk), ds.chunkSize)
 	}
 	ds.squareRow[x][y] = newChunk
 	ds.squareCol[y][x] = newChunk
 	ds.resetRoots()
+	return nil
 }
 
-// setCell sets a specific cell.
-func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) {
+// setCell sets a specific cell. setCell will overwrite any existing value.
+// Returns an error if the newChunk is not the correct size.
+func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) error {
+	if len(newChunk) != int(ds.chunkSize) {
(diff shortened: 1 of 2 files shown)


evan-forbes (dismissed):

datasquare_test.go
```diff
@@ -53,18 +53,20 @@ func TestInvalidDataSquareCreation(t *testing.T) {
...
-	if err != nil {
-		panic(err)
-	}
+	assert.NoError(t, err)
 
-	// SetCell can only write to nil cells
-	assert.Panics(t, func() { ds.SetCell(0, 0, []byte{0}) })
+	err = ds.setCell(0, 0, []byte{42})
+	assert.NoError(t, err)
+	assert.Equal(t, []byte{42}, ds.GetCell(0, 0))
```
staheri14, line 60: I think it is worth adding a test for `setCell` to add a cell with the wrong size.

datasquare_test.go
```diff
@@ -53,18 +53,20 @@ func TestInvalidDataSquareCreation(t *testing.T) {
...
-	ds.setCell(0, 0, nil)
+	// SetCell can only write nil cells so this should return an error.
+	err = ds.SetCell(0, 0, []byte{0})
+	assert.Error(t, err)
 
-	ds.SetCell(0, 0, []byte{42})
-	assert.Equal(t, []byte{42}, ds.GetCell(0, 0))
+	// SetCell can only write chunks of the same size as the data square so this
+	// should return an error.
+	err = ds.SetCell(0, 0, []byte{0, 0})
```
staheri14, line 68: My understanding is that since the cell at position `0,0` is already filled in previous parts, the error returned here is not due to the wrong size of the `newChunk` (rather the cell being non-empty). That is why the codecov complains about the following uncovered lines `datasquare.go#L293-L294`.

staheri14 (commented):
LGTM, only one fix is needed for the tests.

datasquare_test.go
```diff
@@ -53,18 +53,20 @@ func TestInvalidDataSquareCreation(t *testing.T) {
...
+	assert.Error(t, err)
```
staheri14, line 69: Also, a successful test (with no error) for the `SetCell` will eliminate the other complaints of the codcov.

rootulp:
I just noticed https://github.com/celestiaorg/rsmt2d/issues/127 so tagging @Wondertan for review

rootulp:
Based on https://github.com/celestiaorg/rsmt2d/issues/127#issuecomment-1261079482 we may not want to return an `error` here b/c that's not ergonomic for celestia-node. Open to alternatives

staheri14 (approved):
LGTM!

Wondertan:
@rootulp, returning an error is ok. This PR closes the #127

I forgot to use permalinks in the issue🤦🏻

datasquare.go
```diff
@@ -286,22 +286,31 @@ func (ds *dataSquare) GetCell(x uint, y uint) []byte {
...
+	return nil
 }
 
-// setCell sets a specific cell.
-func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) {
+// setCell sets a specific cell. setCell will overwrite any existing value.
+// Returns an error if the newChunk is not the correct size.
+func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) error {
+	if len(newChunk) != int(ds.chunkSize) {
+		return fmt.Errorf("cannot set cell with chunk size %d because dataSquare chunk size is %d", len(newChunk), ds.chunkSize)
```
Wondertan, line 308: What's the reason for having two setters? I don't remember

rootulp: My guess is the exported `SetCell` shouldn't have access to overwrite populated cells but the internal `setCell` should 🤷‍♂️

Wondertan (approved):

evan-forbes (approved):

Issue #178, opened 2023-06-20
ErrByzantineData.Shares are not filled


andrijamitrovic23:
`ErrByzantineData.Shares` are not set when verification of newly completed orthogonal vectors returns a `ErrByzantineData` error ([first place](https://github.com/celestiaorg/rsmt2d/blob/6515446be623987ded119e530900589f30e4c69e/extendeddatacrossword.go#L164-L167) and [second place](https://github.com/celestiaorg/rsmt2d/blob/6515446be623987ded119e530900589f30e4c69e/extendeddatacrossword.go#L229-L232)). On the contrary, these shares are set if a verification of newly completed [row](https://github.com/celestiaorg/rsmt2d/blob/6515446be623987ded119e530900589f30e4c69e/extendeddatacrossword.go#L152-L152) or [column](https://github.com/celestiaorg/rsmt2d/blob/6515446be623987ded119e530900589f30e4c69e/extendeddatacrossword.go#L217) returns `ErrByzantineData` error.

[...]

PR #190, opened 2023-06-21
fix: populate shares in `ErrByzantineData`

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/178

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -163,6 +163,10 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 		if noMissingData(col, r) { // not completed
 			err := eds.verifyAgainstColRoots(colRoots, uint(c), col, r, rebuiltShares[c])
 			if err != nil {
+				var byzErr *ErrByzantineData
+				if errors.As(err, &byzErr) {
+					byzErr.Shares = shares
+				}
 				return false, false, err
 			}
 		}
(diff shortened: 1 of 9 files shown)


extendeddatacrossword.go
```diff
@@ -164,6 +164,10 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 		if noMissingData(col, r) { // not completed
```
staheri14, line 164: I think this `if` never gets through, because, despite a new share of this col being rebuilt, yet, that share hasn't been placed in the col, so, the `noMissingData` returns `false`, i.e., it is incomplete. I think we need to update the column before checking if it is complete or not i.e., https://github.com/celestiaorg/rsmt2d/blob/9ed797dd43ffbee9e5212d769f0330419093a123/extendeddatacrossword.go#L165 (it points to the branch I am working on for  https://github.com/celestiaorg/rsmt2d/issues/191). WDYT?

staheri14: > Are you asking if this conditional ever evaluates to true? 

Yes, that is right. 
BTW, I found the answer to my question. The `noMissingData` initially did not seem to take into account the newly constructed share when counting the missing shares. However, it actually does where it skips the `rebuiltIndex`  while iterating over the shares of a row/col i.e., [here](https://github.com/celestiaorg/rsmt2d/blob/e8e042fdc8dcf485510c093a9c831f5b2fac2d3b/extendeddatacrossword.go#L397).

staheri14 (approved):
LGTM! also liked your test documentation 👍

evan-forbes (approved):
nice simple solution LGTM!

PR #195, opened 2023-06-23
chore!: export codec methods

evan-forbes:
## Overview

If the methods for the `Codec` interface are exported, then users can define their own interfaces outside of this package. I'm not sure if leaving the methods unexported was done on purpose, so if it was this PR can be closed.

edit: for posterity this [PR](https://github.com/celestiaorg/rsmt2d/pull/46/files) exported only a few methods, so it looks like it wasn't done on purpose, but I'm still unsure if there are any other specific reasons not to 

## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

### DIFF
--- a/codecs.go
+++ b/codecs.go
@@ -18,10 +18,10 @@ type Codec interface {
 	// Decode decodes sparse original + parity data, automatically extracting share size.
 	// Missing shares must be nil. Returns original + parity data.
 	Decode(data [][]byte) ([][]byte, error)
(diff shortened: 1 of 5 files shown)


rootulp (approved):

cmwaters (approved):
Makes sense to me

staheri14 (approved):

Issue #196, opened 2023-06-26
Why rebuild extended shares?


rootulp:
## Context

https://github.com/celestiaorg/rsmt2d/blob/328fe4dba32ab858b06e60fde9f12252f7938450/extendeddatacrossword.go#L244-L275

## Question

Given that `Decode` returns the original shares + parity shares, why does `rebuildShares` have logic to separately rebuild extended shares? It seems possible to entirely remove the `isExtendedPartIncomplete` conditional.

adlerjohn:
IIRC, Infectious behaves differently and `Decode` doesn't return all shares. Could you confirm?

rootulp:
Ahh I see, thanks for pointing this out. It appears Infectious `Decode` did only return the rebuilt shares [here](https://github.com/celestiaorg/rsmt2d/pull/154/files#diff-b2ca25b4b25855e5d483204f9dfc73df2e2e2e3946cebafa0aed2e84dfb37e6aL74).

adlerjohn:
This should probably have been originally fixed within the Infectious codec IMO, rather that doubling work for other codecs.

PR #197, opened 2023-06-26
refactor: simplify `rebuildShares`

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/196

Since [`Decode`](https://github.com/rootulp/rsmt2d/blob/1f1904acc114b41dff838ad1d530feacc1d9f199/leopard.go#L46-L54) returns original + parity shares, there is no need to separately rebuild the parity shares. This PR removes an unnecessary conditional inside `rebuildShares` which let us also remove a param to that function. After the refactor, a few helper methods were no longer used so they were also removed.

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -259,25 +259,6 @@ func (eds *ExtendedDataSquare) rebuildShares(
 		return nil, false, nil
 	}
 
-	if isExtendedPartIncomplete {
-		// If needed, rebuild the parity shares too.
-		rebuiltExtendedShares, err := eds.codec.Encode(rebuiltShares[0:eds.originalDataWidth])
-		if err != nil {
-			return nil, true, err
-		}
-		rebuiltShares = append(
-			rebuiltShares[0:eds.originalDataWidth],
-			rebuiltExtendedShares...,
-		)
-	} else {
-		// Otherwise copy them from the EDS.
-		startIndex := len(shares) - int(eds.originalDataWidth)
-		rebuiltShares = append(
-			rebuiltShares[0:eds.originalDataWidth],
-			shares[startIndex:]...,
-		)
-	}
-
 	return rebuiltShares, true, nil
 }
 
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -136,9 +136,8 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 		shares[c] = vectorData[c]
 	}
 
-	isExtendedPartIncomplete := !eds.rowRangeNoMissingData(uint(r), eds.originalDataWidth, eds.width)
 	// Attempt rebuild
-	rebuiltShares, isDecoded, err := eds.rebuildShares(isExtendedPartIncomplete, shares)
+	rebuiltShares, isDecoded, err := eds.rebuildShares(shares)
 	if err != nil {
 		return false, false, err
 	}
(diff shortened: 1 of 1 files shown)


extendeddatacrossword.go
```diff
@@ -249,7 +247,6 @@ func (eds *ExtendedDataSquare) solveCrosswordCol(
 // 2. Whether the original shares could be decoded from the shares parameter.
 // 3. [Optional] an error.
 func (eds *ExtendedDataSquare) rebuildShares(
```
evan-forbes, line 249: do we need this function at all if its just decoding?

evan-forbes (approved):
ahh okay so the encode and decode abstractions were (are?) leaky

extendeddatacrossword.go
```diff
@@ -259,25 +256,6 @@ func (eds *ExtendedDataSquare) rebuildShares(
 		return nil, false, nil
 	}
 
-	if isExtendedPartIncomplete {
```
staheri14, line 262: [Question] I suppose there have not been any tests associated with this part of the code, as no test file has been updated in this PR, right?

extendeddatacrossword.go
```diff
@@ -259,25 +256,6 @@ func (eds *ExtendedDataSquare) rebuildShares(
...
-		// If needed, rebuild the parity shares too.
-		rebuiltExtendedShares, err := eds.codec.Encode(rebuiltShares[0:eds.originalDataWidth])
```
staheri14, line 264: [Question] The only potential issue could arise if the parity portion is incomplete and the rebuilt extended shares obtained from `rebuiltShares, err := eds.codec.Decode(shares)` (the second half of  `rebuiltShares`) and `rebuiltExtendedShares, err := eds.codec.Encode(rebuiltShares[0:eds.originalDataWidth])` turn out to be different. Have we conducted any tests to verify this?

In the original code, we initially decode the entire row, and if the extended part (before decoding) is incomplete, we discard that part from the decoded version and replace it with the encoding result. However, in the new version, we retain whatever the decode function provides us with. If we are confident that these two have the same result, then everything looks good to me! (I think they should be the same)

staheri14 (approved):
LGTM! left two questions, but they are non-blocking as long as we are certain about them.

extendeddatacrossword.go
rootulp: good question, I can explore deleting it entirely.

extendeddatacrossword.go
rootulp: codecov claims these lines were covered: https://app.codecov.io/gh/celestiaorg/rsmt2d/blob/master/extendeddatacrossword.go

I assume via https://github.com/rootulp/rsmt2d/blob/1f1904acc114b41dff838ad1d530feacc1d9f199/extendeddatacrossword_test.go#L21

extendeddatacrossword.go
rootulp: > The only potential issue could arise if the parity portion is incomplete

I don't expect this to happen because Decode calls Reconstruct [here](https://github.com/rootulp/rsmt2d/blob/1f1904acc114b41dff838ad1d530feacc1d9f199/leopard.go#L52) and the godoc for Reconstruct claims:

> 	// If there are too few shards to reconstruct the missing
	// ones, ErrTooFewShards will be returned.

Ref: https://github.com/klauspost/reedsolomon/blob/523164698be98f1603cf1235f5a1de17728b2091/reedsolomon.go#L56-L57

extendeddatacrossword.go
rootulp: TBH I'm only creating a new issue rather than tackling in this PR b/c I don't want to dismiss existing approvals 😞 https://github.com/celestiaorg/rsmt2d/issues/199

Issue #202, opened 2023-06-30
Return value of `setCell` isn't checked


rootulp:
## Context

https://github.com/celestiaorg/rsmt2d/blob/25576208f90d3d171e5cbf7407704c1d61ae6bf1/extendeddatacrossword_test.go#L209

## Problem

After https://github.com/celestiaorg/rsmt2d/pull/172 `setCell` may return an error and indeed does when trying to overwrite a cell with `nil`. This means an error is going undetected in the test.

### Option A

Revert  `setCell` to not perform input validation.

### Option B

Check the the return value of `setCell` in the test. Fail the test if an error is observed.

### Follow ups

https://github.com/celestiaorg/rsmt2d/issues/204

rootulp:
Based on the tests in [extendeddatacrossword_test.go](https://github.com/celestiaorg/rsmt2d/blob/25576208f90d3d171e5cbf7407704c1d61ae6bf1/extendeddatacrossword_test.go#L209), the desired behavior is that `setCell` is able to overwrite cells with `nil` so going with **Option A**.

PR #203, opened 2023-06-30
test: do not perform input validation in `setCell`

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/202 via **Option A**

Note: this isn't breaking b/c `setCell` isn't exported.

### DIFF
--- a/datasquare.go
+++ b/datasquare.go
@@ -307,15 +307,6 @@ func (ds *dataSquare) SetCell(x uint, y uint, newChunk []byte) error {
 	return nil
 }
 
-// setCell overwrites the contents of a specific cell. setCell does not perform
-// any input validation so most usecases should use `SetCell` instead of
-// `setCell`. This method exists strictly for testing.
-func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) {
-	ds.squareRow[x][y] = newChunk
-	ds.squareCol[y][x] = newChunk
-	ds.resetRoots()
-}
-
 // Flattened returns the concatenated rows of the data square.
 func (ds *dataSquare) Flattened() [][]byte {
 	flattened := [][]byte(nil)
--- a/datasquare.go
+++ b/datasquare.go
@@ -307,16 +307,13 @@ func (ds *dataSquare) SetCell(x uint, y uint, newChunk []byte) error {
 	return nil
 }
 
-// setCell sets a specific cell. setCell will overwrite any existing value.
-// Returns an error if the newChunk is not the correct size.
-func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) error {
-	if len(newChunk) != int(ds.chunkSize) {
-		return fmt.Errorf("cannot set cell with chunk size %d because dataSquare chunk size is %d", len(newChunk), ds.chunkSize)
-	}
+// setCell overwrites the contents of a specific cell. setCell does not perform
+// any input validation so most usecases should use `SetCell` instead of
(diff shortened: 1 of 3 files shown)


datasquare.go
```diff
@@ -307,16 +307,13 @@ func (ds *dataSquare) SetCell(x uint, y uint, newChunk []byte) error {
...
-// setCell sets a specific cell. setCell will overwrite any existing value.
-// Returns an error if the newChunk is not the correct size.
-func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) error {
-	if len(newChunk) != int(ds.chunkSize) {
-		return fmt.Errorf("cannot set cell with chunk size %d because dataSquare chunk size is %d", len(newChunk), ds.chunkSize)
-	}
+// setCell overwrites the contents of a specific cell. setCell does not perform
+// any input validation so most usecases should use `SetCell` instead of
+// `setCell`. This method exists strictly for testing.
+func (ds *dataSquare) setCell(x uint, y uint, newChunk []byte) {
```
staheri14, line 313: Since this is meant to be used for testing purposes, I think it's a good idea to move it to a test package. That way, we can avoid any accidental misuse.

staheri14:
One alternative  solution could be to call `SetCell` in the test only for non-nill values, i.e., in [here](https://github.com/celestiaorg/rsmt2d/blob/25576208f90d3d171e5cbf7407704c1d61ae6bf1/extendeddatacrossword_test.go#L209).
```go
  for i, coords := range test.coords {
          x := coords[0]
          y := coords[1]
          if test.values[i]!= nil {
             eds.setCell(x, y, test.values[i])
          }
  }
```

datasquare.go
staheri14: One alternative  solution could be to call `SetCell` in the test only for non-nill values, i.e., in [here](https://github.com/celestiaorg/rsmt2d/blob/25576208f90d3d171e5cbf7407704c1d61ae6bf1/extendeddatacrossword_test.go#L209).
```go
  for i, coords := range test.coords {
          x := coords[0]
          y := coords[1]
          if test.values[i]!= nil {
             eds.setCell(x, y, test.values[i])
          }
  }
```

rootulp: Hmm I don't understand the proposed alternative solution. The test wants to overwrite nil and non-nil values. Do you mind rephrasing? Also are you referring to `setCell` or `SetCell`?

rootulp: > I think it's a good idea to move it to a test package

Can do!

staheri14: > Hmm I don't understand the proposed alternative solution. The test wants to overwrite nil and non-nil values. Do you mind rephrasing? Also are you referring to `setCell` or `SetCell`?

Oh, I mistakenly assumed that the `eds` contains `nil` cells by default so the tests are can be reduced to altering nil cells to non-nil cells, but now I see that `eds` initially gets populated with non-nil values in `createTestEds`, so my alternative solution does not apply in here.

staheri14 (approved):
Thanks for addressing the comments, LGTM!

datasquare_test.go
```diff
@@ -95,47 +95,40 @@ func TestSetCell(t *testing.T) {
 	}
 }
 
+// Test_setCell verifies that setCell can overwrite cells without performing any
+// input validation.
 func Test_setCell(t *testing.T) {
```
rootulp, line 100: on re-review it seems overkill to have a test for an un-exported function that only exists for the sake of testing so I can remove this unit test if reviewers request

Issue #204, opened 2023-06-30
Enable errcheck linter


rootulp:
## Motivation

https://github.com/celestiaorg/rsmt2d/issues/202

## Proposal

Run golangci-lint and specifically [errcheck](https://github.com/kisielk/errcheck) (enabled by default on golangci-lint)

PR #208, opened 2023-07-03
feat!: enable golangci-lint

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/204

Marked as breaking b/c the function signature for exported functions (`RowRoots` and `ColRoots`) was modified to include an error return parameter.

## Testing

1. `golangci-lint run` passes locally
2. `golangci-lint run` passes in CI [here](https://github.com/celestiaorg/rsmt2d/actions/runs/5456354873/jobs/9929053344?pr=208#step:4:27)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,28 @@
+# lint runs golangci-lint
+name: lint
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - release/**
+
+env:
+  GO_VERSION: '1.20'
+
+jobs:
+  golangci-lint:
+    name: golangci-lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 8
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - uses: golangci/golangci-lint-action@v3.6.0
+        with:
+          version: v1.52.2
+          args: --timeout 10m
+          github-token: ${{ secrets.github_token }}
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,17 @@
+run:
+  timeout: 5m
+  modules-download-mode: readonly
+
+linters:
+  enable:
+    - exportloopref
+    - gofumpt
+    - misspell
+    - nakedret
+    - revive
+    - prealloc
+
+linters-settings:
+  nakedret:
+    # Ban the use of naked returns because they reduce code readability.
+    max-func-lines: 0 # override the default: 30
--- a/README.md
+++ b/README.md
@@ -71,10 +71,20 @@ func main() {
 }
 ```
 
-## Building From Source
+## Contributing
 
-Run benchmarks
+1. [Install Go](https://go.dev/doc/install) 1.19+
+1. [Install golangci-lint](https://golangci-lint.run/usage/install/)
+
+### Helpful Commands
 
 ```sh
+# Run unit tests
+go test ./...
+
+# Run benchmarks
 go test -benchmem -bench=.
+
+# Run linter
+golangci-lint run
 ```
--- a/datasquare_test.go
+++ b/datasquare_test.go
@@ -337,10 +337,9 @@ func Test_setRowSlice(t *testing.T) {
 		if tc.wantErr {
 			assert.Error(t, err)
 			return
-		} else {
-			assert.NoError(t, err)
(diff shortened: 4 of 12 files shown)


.golangci.yml
```diff
@@ -0,0 +1,17 @@
...
+  modules-download-mode: readonly
+
+linters:
+  enable:
+    - exportloopref
+    - gofumpt
+    - misspell
+    - nakedret
+    - revive
+    - prealloc
```
rootulp, line 12: copied the celestia-app `.golanci.yml` b/c that seems like a good starting point / uncontroversial

extendeddatacrossword.go
```diff
@@ -178,7 +178,13 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 
 	// Insert rebuilt shares into square.
 	for c, s := range rebuiltShares {
-		eds.SetCell(uint(r), uint(c), s)
```
rootulp, line 181: previously this was silently ignoring an error! Thanks to errcheck, we are forced to check the error return type and propagate it which resulted in:

```
--- FAIL: TestRepairExtendedDataSquare (0.00s)
    --- FAIL: TestRepairExtendedDataSquare/leopard (0.00s)
        --- FAIL: TestRepairExtendedDataSquare/leopard/MaximumErasures (0.00s)
            /Users/rootulp/git/rootulp/celestia/rsmt2d/extendeddatacrossword_test.go:60: unexpected err while repairing data square: cannot set cell (3, 2) as it already has a value 05050505050505050505050505050505050505050505050505050505050505050505050505050505050505050505050505050505050505050505050505050505, codec: :leopard
```

To resolve, 2b4078bd7d01f8cbf125be4c533d09cb33c55072 checks if the cellToSet is nil prior to the attempt to `SetCell`

.github/workflows/lint.yml
```diff
@@ -0,0 +1,28 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - release/**
+
+env:
+  GO_VERSION: '1.20'
```
rootulp, line 12: inconsistent until https://github.com/celestiaorg/rsmt2d/pull/211 merges

datasquare_test.go
```diff
@@ -423,7 +423,10 @@ func computeRowProof(ds *dataSquare, x uint, y uint) ([]byte, [][]byte, uint, ui
 	data := ds.row(x)
 
 	for i := uint(0); i < ds.width; i++ {
-		tree.Push(data[i])
+		err := tree.Push(data[i])
+		if err != nil {
+			return nil, nil, 0, 0, err
```
staheri14, line 428: [Suggestion] I'd suggest adding a godoc for this function `computeRowProof`.  It makes it easier to interpret the outputs.

datasquare_test.go
```diff
@@ -445,7 +448,7 @@ type errorTree struct {
 	leaves [][]byte
 }
 
-func newErrorTree(axis Axis, index uint) Tree {
+func newErrorTree(_ Axis, _ uint) Tree {
```
staheri14, line 451: [Question] Why don't we remove the inputs if they are used at all?

staheri14 (approved):
LGTM!

datasquare_test.go
rootulp: +1 I agree. TBH I like named return parameters because they help document the function. Ex:

```go
func computeRowProof(ds *dataSquare, x uint, y uint) (merkleRoot []byte, proof [][]byte, proofIndex uint, numLeaves uint, err error) {
```

but our team has mixed opinions on it so I'll create a follow up issue to add godoc for this.

rootulp: Created https://github.com/celestiaorg/rsmt2d/issues/212

datasquare_test.go
rootulp: Good question. I just attempted to but it looks like this function needs to conform to this type: https://github.com/rootulp/rsmt2d/blob/29390107470a83ae854114d2723993ecd0a01f69/tree.go#L9-L11

evan-forbes (approved):

README.md
```diff
@@ -71,10 +71,20 @@ func main() {
 }
 ```
 
-## Building From Source
+## Contributing
 
-Run benchmarks
+1. [Install Go](https://go.dev/doc/install) 1.19+
```
rootulp, line 76: 🤦‍♂️ this is incorrect after I created https://github.com/celestiaorg/rsmt2d/pull/211

Issue #212, opened 2023-07-04
[Suggestion] I'd suggest adding a godoc for this function `computeRowProof`.  It makes it easier to interpret the outputs.


rootulp:
[Suggestion] I'd suggest adding a godoc for this function `computeRowProof`.  It makes it easier to interpret the outputs.

_Originally posted by @staheri14 in https://github.com/celestiaorg/rsmt2d/pull/208#discussion_r1252233065_

PR #228, opened 2023-07-05
docs: godoc for computeRowProof

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/212

### DIFF
--- a/datasquare_test.go
+++ b/datasquare_test.go
@@ -418,6 +418,9 @@ func BenchmarkEDSRoots(b *testing.B) {
 	}
 }
 
+// computeRowProof returns the merkle root, proof, proofIndex, and numLeaves for
+// the share at coordinates (x, y). Returns an error if the shares in the row
+// fail to push to the Merkle tree.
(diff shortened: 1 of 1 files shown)


datasquare_test.go
```diff
@@ -418,6 +418,9 @@ func BenchmarkEDSRoots(b *testing.B) {
 	}
 }
 
+// computeRowProof returns the merkle root, proof, proofIndex, and numLeaves for
```
staheri14, line 421: It's not immediately clear as what "proof" and "proof index" mean or refer to. It would be great to explain them further.

evan-forbes (dismissed):

datasquare_test.go
rootulp: I just read the implementation in https://github.com/celestiaorg/merkletree/blob/c33506a7aa26be568fbb9714b3a6f66c1b60d2aa/tree.go#L95-L99 and TBH I don't understand it so closing this PR as unresolved.

Issue #220, opened 2023-07-05
Clarification around `rowRoots` provided to `Repair`


rootulp:
## Context

https://github.com/celestiaorg/rsmt2d/blob/e8e042fdc8dcf485510c093a9c831f5b2fac2d3b/extendeddatacrossword.go#L59-L74

## Question

Should the `rowRoots` and `columnRoots` that are provided to `Repair` be the row roots that are computed on the complete  EDS or the incomplete EDS? I assume the complete EDS because computing row roots for an incomplete EDS is undefined behavior (see https://github.com/celestiaorg/rsmt2d/issues/122).

However, if the row roots are indeed the ones computed on the complete EDS prior to shares being removed, I don't understand why [this test](https://github.com/celestiaorg/rsmt2d/blob/e8e042fdc8dcf485510c093a9c831f5b2fac2d3b/extendeddatacrossword_test.go#L186-L243) exists. I would expect every test case in that test to fail because `Repair` is being called with bad roots. The only reason test cases in that test pass is because row roots are being computed _after_ shares are removed from the data square which is undefined behavior.

staheri14:
> However, if the row roots are indeed the ones computed on the complete EDS prior to shares being removed, I don't understand why [this test](https://github.com/celestiaorg/rsmt2d/blob/e8e042fdc8dcf485510c093a9c831f5b2fac2d3b/extendeddatacrossword_test.go#L186-L243) exists. I would expect every test case in that test to fail because Repair is being called with bad roots. The only reason test cases in that test pass is because row roots are being computed after shares are removed from the data square which is undefined behavior.


Interestingly, I also contemplated the same issue yesterday regarding why roots are computed from the corrupted EDS. As mentioned by @Wondertan, the roots should be derived from the DA header, which, in the tests, represents the roots of the sample EDS before it got corrupted.

Regarding the specific [test](https://github.com/celestiaorg/rsmt2d/blob/e8e042fdc8dcf485510c093a9c831f5b2fac2d3b/extendeddatacrossword_test.go#L186-L243), my understanding is that it seems to aim at examining whether a full node can identify faulty erasure coding, even when the roots of the DA header are corrupted, meaning they were computed from a corrupted EDS.
However, even if this is the scneario, I am still unsure why one can even compute a row root while some shares are nil (as also mentioned in this issue https://github.com/celestiaorg/rsmt2d/issues/214)

rootulp:
[...]

PR #240, opened 2023-07-07
docs: clarify rowRoots to Repair

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/220

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -56,13 +56,12 @@ func (e *ErrByzantineData) Error() string {
 	return fmt.Sprintf("byzantine %s: %d", e.Axis, e.Index)
 }
 
-// Repair attempts to repair an incomplete extended data
-// square (EDS), comparing repaired rows and columns against expected Merkle
-// roots.
-//
-// # Input
-//
-// Missing shares must be nil.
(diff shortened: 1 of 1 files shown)


extendeddatacrossword.go
```diff
@@ -56,13 +56,12 @@ func (e *ErrByzantineData) Error() string {
...
-// square (EDS), comparing repaired rows and columns against expected Merkle
-// roots.
-//
-// # Input
-//
-// Missing shares must be nil.
+// Repair attempts to repair an incomplete extended data square (EDS). The
+// parameters rowRoots and colRoots are the expected Merkle roots for each row
+// and column. rowRoots and colRoots are used to verify that a repaired row or
+// column is correct. Prior to the repair process, if a row or column is already
```
cmwaters, line 62: Q: you can technically repair an EDS without the rowRoots and colRoots right?

cmwaters (approved):
LGTM

extendeddatacrossword.go
rootulp: Yes it is possible because the underlying library (klauspost/reedsolomon) doesn't have a notion of verifying the repaired data against a row root or col root.

Issue #237, opened 2023-07-07
Update chunkSize in tests to `512`


rootulp:
this appears to have been set to share size, if that's the case then perhaps we should update these to 512

_Originally posted by @evan-forbes in https://github.com/celestiaorg/rsmt2d/pull/236#discussion_r1254992505_

rootulp:
To elaborate: the share size in celestia-app has increased from 256 to 512 so the tests in rsmt2d could more accurately reflect the usage in celestia-app by increasing to 512.

Issue #216, opened 2023-07-05
Remove test cases that vary based on codec for extendeddatacrossword_test.go


rootulp:
## Context

The tests in [extendeddatacrossword_test.go](https://github.com/celestiaorg/rsmt2d/blob/e8e042fdc8dcf485510c093a9c831f5b2fac2d3b/extendeddatacrossword_test.go) contain test cases for multiple codecs but after https://github.com/celestiaorg/rsmt2d/issues/151 this repo only supports one codec.

Example: https://github.com/celestiaorg/rsmt2d/blob/e8e042fdc8dcf485510c093a9c831f5b2fac2d3b/extendeddatacrossword_test.go#L24-L31

## Proposal

Refactor the tests in this file to remove test cases that differ based on codec. This will likely remove a level of indentation from the interesting parts of this file.

rootulp:
I don't see any GH issues in this repo for adding new codecs. I think the current test structure is a byproduct of development (i.e. historically multiple codecs were supported) so IMO we should refactor to just support the one codec this repo supports. 

In the future, if we need to add a new codec, we'll need test coverage of the new codec so the PR that adds a new codec and refactor to add test cases for it.

PR #242, opened 2023-07-07
feat: extends error handling to detect byzantine data and unordered shares in tree construction

### LINKED ISSUE #191: Handle `out of order` shares in the data square
Based on [this](https://github.com/celestiaorg/celestia-node/issues/2377) we may run in a situation when shares will be out of order in the dataSquare. In this case, we will not be able to compute the Merkle root of row or col and reconstruction fails. To handle such error properly we need to get these shares+indexes of Merkel Roots of the opposite axis(to build a Merkle Proof against them). If Merkle Roots for both row/col can't be computed, then we should receive an error that data is not available.

staheri14:
## Overview
Closes #191 

## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Visual proof for any user facing features like CLI or documentation updates
- [x] Linked issues closed with keywords


extendeddatacrossword_test.go
```diff
@@ -344,3 +345,143 @@ func createTestEds(codec Codec, shareSize int) *ExtendedDataSquare {
 
 	return eds
 }
+
+func TestCorruptedEdsReturnsErrByzantineData_UorderedShares(t *testing.T) {
```
rootulp, line 349: typo
```suggestion
func TestCorruptedEdsReturnsErrByzantineData_UnorderedShares(t *testing.T) {
```

extendeddatacrossword_test.go
```diff
@@ -344,3 +345,143 @@ func createTestEds(codec Codec, shareSize int) *ExtendedDataSquare {
...
+	shareSize := 64
```
rootulp, line 350: [optional]

```suggestion
	shareSize := 512
```

motivation: https://github.com/celestiaorg/rsmt2d/issues/237

extendeddatacrossword_test.go
```diff
@@ -344,3 +345,143 @@ func createTestEds(codec Codec, shareSize int) *ExtendedDataSquare {
...
+	two := bytes.Repeat([]byte{2}, shareSize)
+	three := bytes.Repeat([]byte{3}, shareSize)
+	sharesValue := []int{1, 2, 3, 4}
+	tests := []struct {
+		name          string
+		coords        [][]uint
+		values        [][]byte
+		wantErr       bool
+		corruptedAxis Axis
+		corruptedInd  uint
```
rootulp, line 362: [nit] it wasn't immediately obvious what `Ind` meant but it appears short for index. This suggestion can't be applied directly b/c it implies renaming the variable in this scope

```suggestion
		corruptedIndex  uint
```

extendeddatacrossword_test.go
```diff
@@ -344,3 +345,143 @@ func createTestEds(codec Codec, shareSize int) *ExtendedDataSquare {
...
+				three, nil, nil, nil,
+				one, nil, nil, nil,
+				nil, nil, nil,
+				nil, nil, nil,
+			},
+			corruptedInd: 0,
+		},
+	}
+
+	for codecName, codec := range codecs {
```
rootulp, line 428: [optional] there's only one codec so we can remove a layer of nesting from this test case

See: https://github.com/celestiaorg/rsmt2d/issues/216

extendeddatacrossword_test.go
```diff
@@ -344,3 +345,143 @@ func createTestEds(codec Codec, shareSize int) *ExtendedDataSquare {
...
+			dAHeaderRoots, err := eds.getRowRoots()
+			assert.NoError(t, err)
+
+			dAHeaderCols, err := eds.getColRoots()
+			assert.NoError(t, err)
+			for _, test := range tests {
+				t.Run(test.name, func(t *testing.T) {
+					// create an eds with the given shares
+					corruptEds := createTestEdsWithNMT(t, codec, shareSize, namespaceSize, sharesValue...)
+					assert.NotNil(t, eds)
```
rootulp, line 442: [question] this line appears redundant with

```go
			eds := createTestEdsWithNMT(t, codec, shareSize, namespaceSize, 1, 2, 3, 4)
			assert.NotNil(t, eds)
```

from above. Does it mean to do

```suggestion
					assert.NotNil(t, corruptEds)
```

extendeddatacrossword_test.go
```diff
@@ -344,3 +345,143 @@ func createTestEds(codec Codec, shareSize int) *ExtendedDataSquare {
...
+	for i, shareValue := range sharesValue {
+		shares[i] = bytes.Repeat([]byte{byte(shareValue)}, shareSize)
+	}
+	edsWidth := 4            // number of shares per row/column in the extended data square
+	odsWidth := edsWidth / 2 // number of shares per row/column in the original data square
+
+	eds, err := ComputeExtendedDataSquare(shares, codec, newConstructor(uint64(odsWidth), nmt.NamespaceIDSize(namespaceSize)))
+	if err != nil {
+		panic(err)
+	}
```
rootulp, line 484: [optional] this can be replaced with

```suggestion
	require.NoError(t, err)
```

nmtwrapper.go
```diff
@@ -0,0 +1,142 @@
+package rsmt2d
+
+// The contents of this file have been adapted from the source file available at https://github.com/celestiaorg/celestia-app/blob/main/pkg/wrapper/nmt_wrapper.go,
```
rootulp, line 3: [question][not blocking] should we use a permalink in case this file moves? I don't have a strong preference

rootulp (dismissed):
No blocking feedback. Question about the copied file `nmtwrapper.go`: is it only needed for tests? If yes, is it possible to use the `_test.go` suffix?

nmtwrapper.go
evan-forbes, line 3: the only thing the wrapper imports is the [consts](https://github.com/celestiaorg/celestia-app/blob/5bcf4ad51756e5f31dd7d9ea140844063b6ce391/pkg/wrapper/nmt_wrapper.go#L6C53-L6C53), so maybe it makes sense to accept those consts as args, then move the entire wrapper here

evan-forbes (dismissed):
LGTM :+1:

nmtwrapper.go
Wondertan, line 3: Why copy and why not have these tests next to the wrapper?

Wondertan: Or why not move the wrapper here completely then?

Wondertan (approved):

extendeddatacrossword_test.go
staheri14: Done [2dcfb41](https://github.com/celestiaorg/rsmt2d/pull/242/commits/2dcfb41c7bd30da38fafad1ec303a62bc9b6dfc9).

extendeddatacrossword_test.go
staheri14: Right, it was meant to test whether `corruptedEds` is nil or not. Fixed in https://github.com/celestiaorg/rsmt2d/pull/242/commits/2dcfb41c7bd30da38fafad1ec303a62bc9b6dfc9.

extendeddatacrossword_test.go
staheri14: Sure, done https://github.com/celestiaorg/rsmt2d/pull/242/commits/2dcfb41c7bd30da38fafad1ec303a62bc9b6dfc9.

nmtwrapper.go
staheri14: Makes sense to me.  Revised the link https://github.com/celestiaorg/rsmt2d/pull/242/commits/a364d791b546ce08e8f9da5fa926bcd53a0e8f33.

extendeddatacrossword_test.go
staheri14: Thanks for mentioning that! I totally agree with eliminating that nested loop. But to keep our PRs more focused and organized, I'd suggest saving this change for a separate PR. Especially since we will have an upcoming PR specifically targeting that issue.

nmtwrapper.go
staheri14: Thanks for the suggestion, here is the tracking issue https://github.com/celestiaorg/rsmt2d/issues/248.

staheri14: > Or why not move the wrapper here completely then?

That also makes sense to me, cannot see any immediate issue for doing so, however, since it requires some refactoring in the code, I'd say it merits a separate PR. Please see the tracking issue https://github.com/celestiaorg/rsmt2d/issues/248.

staheri14: That also makes sense to me, cannot see any immediate issue for doing so, however, since it requires some refactoring in the code, I'd say it merits a separate PR. Please see the tracking issue https://github.com/celestiaorg/rsmt2d/issues/248.

extendeddatacrossword_test.go
staheri14: > This suggestion can't be applied directly b/c it implies renaming the variable in this scope

Can you please expand on this second part of your comment? Nevertheless, I applied the first part of your feedback in https://github.com/celestiaorg/rsmt2d/pull/242/commits/1a4e055a9427c3e37e28dc72d22922d4dc754443

staheri14:
> No blocking feedback. Question about the copied file `nmtwrapper.go`: is it only needed for tests? If yes, is it possible to use the `_test.go` suffix

Indeed, it serves solely for testing purposes, and I have implemented your suggestion by adding the `_test` suffix.  Please see https://github.com/celestiaorg/rsmt2d/pull/242/commits/799ee73e2a8c8e4a26b3f7e2b1cebfb835ace15b.

rootulp (approved):
Still LGTM

nmtwrapper_test.go
```diff
@@ -1,6 +1,6 @@
 package rsmt2d
 
-// The contents of this file have been adapted from the source file available at https://github.com/celestiaorg/celestia-app/blob/main/pkg/wrapper/nmt_wrapper.go,
+// The contents of this file have been adapted from the source file available at https://github.com/celestiaorg/celestia-app/blob/bab6c0d0befe677ab8c2f4b83561c08affc7203e/pkg/wrapper/nmt_wrapper.go#L1,
```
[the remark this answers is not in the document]
staheri14: Thanks, fixed it in https://github.com/celestiaorg/rsmt2d/pull/242/commits/1e7b68f01e58ab4a5078841d4ff016e64f7f94c2.

rootulp (approved):

evan-forbes (approved):

Issue #235, opened 2023-07-06
enforce that chunkSize is a multiple of 64


rootulp:
## Context

https://github.com/rootulp/rsmt2d/blob/5661af1b27d576d5d8cf895b70230e6da4d51fb4/extendeddatasquare_test.go#L40-L43

## Problem

I've observed that chunkSize must be a multiple of 64 but it's not clear to me if this invariant is enforced when importing an extended data square or computing an extended data square.

## Proposal

Investigate if the invariant is enforced. I expect that rsmt2d is propagating errors from klauspost/reedsolomon if shards  aren't of size 64. If that's the case, we could consider introducing a new error and verifying the `data` input for chunk size.

PR #246, opened 2023-07-07
feat: validate chunk size is a multiple of 64 bytes

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/235

### DIFF
--- a/codecs.go
+++ b/codecs.go
@@ -23,6 +23,9 @@ type Codec interface {
 	MaxChunks() int
 	// Name returns the name of the codec.
 	Name() string
+	// ValidateChunkSize returns an error if this codec does not support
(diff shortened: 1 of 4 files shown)


evan-forbes (approved):

cmwaters (approved):
If we expect codecs to only support one chunk size (i.e. 64 in this case), wouldn't be better if the interface was instead `ChunkSize() int` and we compared that value with the chunk size provided.

rootulp:
> If we expect codecs to only support one chunk size (i.e. 64 in this case), wouldn't be better if the interface was instead ChunkSize() int and we compared that value with the chunk size provided.

We expect codecs to support more than one chunk size. In this case, leopard supports chunks of size `64`, `128`, `192`, etc.

Issue #249, opened 2023-07-12
Consider EDS  helper `Roots`


rootulp:
## Context

https://github.com/celestiaorg/celestia-node/pull/2458/files#diff-c33387bdad1a81ed3928f505eadee22744b6fca98b897cacbcb918402843a4cbR246

## Proposal

Consider adding a `Roots()` helper to EDS. Something like:

```go
// Roots returns a byte slice with this eds's RowRoots and ColRoots
// concatenated.
func (eds *ExtendedDataSquare) Roots() (roots [][]byte, err error) {
	rowRoots, err := eds.RowRoots()
	if err != nil {
		return nil, err
	}
	colRoots, err := eds.ColRoots()
	if err != nil {
		return nil, err
	}

	roots = make([][]byte, 0, len(rowRoots)+len(colRoots))
	roots = append(roots, rowRoots...)
	roots = append(roots, colRoots...)
	return roots, nil
}
```

bao1029p:
Hi guys, can we work on this ?

rootulp:
Sure go for it 👍

bao1029p:
@rootulp hey i created a PR can you give approval for the CI to run

PR #270, opened 2023-10-31
feat: add roots helper function

bao1029p:
## Overview

close [#249](https://github.com/celestiaorg/rsmt2d/issues/249)

## Changes

 Add helper function and unit test for new function

### DIFF
--- a/extendeddatasquare_test.go
+++ b/extendeddatasquare_test.go
@@ -426,7 +426,7 @@ func TestEquals(t *testing.T) {
 	})
 }
 
-func TestRootsHelper(t *testing.T) {
+func TestRoots(t *testing.T) {
 	t.Run("returns roots for a 4x4 EDS", func(t *testing.T) {
 		eds, err := ComputeExtendedDataSquare([][]byte{
 			ones, twos,
@@ -435,7 +435,7 @@ func TestRootsHelper(t *testing.T) {
 		require.NoError(t, err)
 
 		roots, err := eds.Roots()
-		assert.NoError(t, err)
+		require.NoError(t, err)
 		assert.Len(t, roots, 8)
 
 		rowRoots, err := eds.RowRoots()
--- a/extendeddatasquare_test.go
+++ b/extendeddatasquare_test.go
@@ -426,6 +426,34 @@ func TestEquals(t *testing.T) {
 	})
 }
 
+func TestRootsHelper(t *testing.T) {
+	t.Run("returns roots for a 4x4 EDS", func(t *testing.T) {
+		eds, err := ComputeExtendedDataSquare([][]byte{
+			ones, twos,
+			threes, fours,
+		}, NewLeoRSCodec(), NewDefaultTree)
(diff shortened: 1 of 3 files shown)


rootulp:
Hey @bao1029p ! Thanks for the PR. I updated the PR title to contain a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) prefix because the PR title will be used as the commit message when we squash and merge this PR.

extendeddatasquare_test.go
```diff
@@ -426,6 +426,34 @@ func TestEquals(t *testing.T) {
 	})
 }
 
+func TestRootsHelper(t *testing.T) {
```
rootulp, line 429: [nit] the `Helper` suffix doesn't seem necessary. Commonly if a function name is `Foo` then the corresponding test will be named `TestFoo`

```suggestion
func TestRoots(t *testing.T) {
```

rootulp: ohh I see why you named this `TestRootsHelper` to avoid a naming collision with the existing `TestRoots`. Okay let's disregard my suggestion for now. In a follow up PR we can rename `TestRoots` => `Test_getRoots` because it's testing the private functions `getRowRoots` and `getColRoots`. Then we can rename this function `TestRoots`.

extendeddatasquare_test.go
```diff
@@ -426,6 +426,34 @@ func TestEquals(t *testing.T) {
...
+	t.Run("returns roots for a 4x4 EDS", func(t *testing.T) {
+		eds, err := ComputeExtendedDataSquare([][]byte{
+			ones, twos,
+			threes, fours,
+		}, NewLeoRSCodec(), NewDefaultTree)
+		require.NoError(t, err)
+
+		roots, err := eds.Roots()
+		assert.NoError(t, err)
+		assert.Len(t, roots, 8)
```
rootulp, line 439: [blocking] this suggestion is a bit exhaustive but we can also verify the exact row roots and column roots. For example:

```go

		rowRoots, err := eds.RowRoots()
		require.NoError(t, err)

		colRoots, err := eds.ColRoots()
		require.NoError(t, err)

		assert.Equal(t, roots[0], rowRoots[0])
		assert.Equal(t, roots[1], rowRoots[1])
		assert.Equal(t, roots[2], rowRoots[2])
		assert.Equal(t, roots[3], rowRoots[3])
		assert.Equal(t, roots[4], colRoots[0])
		assert.Equal(t, roots[5], colRoots[1])
		assert.Equal(t, roots[6], colRoots[2])
		assert.Equal(t, roots[7], colRoots[3])
```

rootulp (commented):
Awesome work! Thanks a ton for the unit test. I have one blocking suggestion on the test and then this LGTM

rootulp (approved):
Thanks for the great work on this @bao1029p !

staheri14 (approved):
Looks good to me too, thanks!

PR #287, opened 2024-01-30
feat!: delete `getTreeNameFromConstructorFn`

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/286 by deleting `getTreeNameFromConstructorFn`. 
Closes https://github.com/celestiaorg/rsmt2d/issues/288, https://github.com/celestiaorg/rsmt2d/issues/289 by fixing those tests.
Inspired by https://github.com/celestiaorg/rsmt2d/pull/278
Includes two breaking changes:
1. `ComputeExtendedDataSquare` now accepts a parameter `treeName: string`
2. `ImportExtendedDataSquare` now accepts a parameter `treeName: string`

PR #277, opened 2023-12-13
[fix] UnmarshalJSON is limited to the default Tree Bug

sontrinh16:
## Overview

Closed: #275 

## Checklist

- [ ] New and updated code has appropriate documentation
- [ ] New and updated code has new and/or updated testing
- [ ] Required CI checks are passing
- [ ] Visual proof for any user facing features like CLI or documentation updates
- [ ] Linked issues closed with keywords

Issue #275, opened 2023-12-11
`UnmarshalJSON` is limited to the default Tree


Wondertan:
## Problem
When we deserialize the EDS [we hardcode the default tree](https://github.com/celestiaorg/rsmt2d/blob/7a8fd8372e84ed73c2e42821b9bb2d0913d18bd1/extendeddatasquare.go#L40C88-L40C88). This is problematic because computed Row and Col roots of the deserialized eds will be utterly different if another custom tree is used.

## Solutions
1. Have a global tree constructor var/registry that `UnmarshalJSON` would use instead
	* Users would then register their trees by name
		* The name would be serialized in the json and deserialization would then get the proper tree by the name
	* This is exactly what happens with the codec.
2.  `UnmarshalEDS` func that accepts the tree.
	* This works in theory, but https://github.com/filecoin-project/go-jsonrpc/blob/17a35772b24f59ceee25cc8d8177a2b255a659ac/client.go#L614 blocks it
	*  If we contribute an option that custom Unmarshall, then we would be able to inject the one for EDS

Wondertan:
The first solution is bulletproof and requires minimum work outside of rsmt2d, as well the same approach is used for Codec, so we should go with it, imo

rootulp:
I don't understand how the original solution was self-describing if celestia-app needs a registry of tree names to tree options. 

I'm in favor of a solution that doesn't need the registry in celestia-app because it was difficult to initialize the registry correctly. IIRC multiple packages in celestia-app were registering the same tree name multiple times.

rootulp:
> Well, self-describing does not imply a lack of registry.

The JSON representation is not self-describing if it depends on a registry from tree name to tree options in order to deserialize. An alternative to writing the tree names in the JSON is to include all of the options needed to describe the tree in the JSON so that no registry is needed. Then the JSON would be self-describing. 

> Do you remember what the difficulty was?
Yes. IIRC multiple packages in celestia-app were registering the same tree name multiple times.

PR #295, opened 2024-02-03
revert!: rsmt2d#277 and rsmt2d#287

rootulp:
While attempting to bump celestia-app to the v0.12.0-rc2, I noticed that the `RegisterTree` design leaks an implementation detail to celestia-app: the registering and managing of `treeName`s. Celestia-app has two categories of of trees: 
1. erasured namespaced merkle tree in [nmt_wrapper.go](https://github.com/celestiaorg/celestia-app/blob/main/pkg/wrapper/nmt_wrapper.go)
2. EDS subtree root cacher [nmt_caching.go](https://github.com/celestiaorg/celestia-app/blob/main/pkg/inclusion/nmt_caching.go)

Each of those categories has trees based on square size and NMT options. Celestia-app needs to be careful to register all the appropriate trees once (and only once) before they are used (via `Compute` or `Import`). I'd like to explore a less breaking option to get celestia-node the original desired feature which was https://github.com/celestiaorg/rsmt2d/issues/275. In the meantime, I think we should revert the two big breaking changes so that main can remain release-able.

Revert https://github.com/celestiaorg/rsmt2d/pull/277
Revert https://github.com/celestiaorg/rsmt2d/pull/287
Closes https://github.com/celestiaorg/rsmt2d/pull/295 because no longer relevant if we merge this.


staheri14:
Mind providing a link to `nmt_caching.go`? Thanks

staheri14:
Has `nmt_caching.go` been utilized in the rsmt2d context? It seems not to conform to the [Tree interface](https://github.com/celestiaorg/rsmt2d/blob/50c7622b937f5da36d4492df55692559211fa822/tree.go#L19-L22), prompting questions about its necessity for registration. Any clarification on this would be greatly appreciated!

rootulp:
I think nmt_caching.go is used in rsmt2d context because [this test](https://github.com/celestiaorg/celestia-app/blob/2d261a2a0f7f627044342929b534a16b762c47e0/pkg/inclusion/nmt_caching_test.go#L120-L122). Also [this line](https://github.com/celestiaorg/celestia-app/blob/22358c30ea3f970048a49aa60b3f7ac977aa0af8/pkg/inclusion/nmt_caching.go#L97) returns a tree that conforms to the rsmt2d Tree interface.

evan-forbes (approved):

staheri14:
Thank you for sharing the links. You're correct; the `NewSubtreeCacher.Constructor` is indeed used in the test, it generates an `ErasuredNamespacedMerkleTree` with caching enabled. It seems, though, that differentiation might be necessary when registering tree names. However, given that the underlying tree remains consistent, do you think there's potential to simplify by consolidating the tree names into a single one? Just pondering this possibility.

Additionally, I'm eager to hear your perspective on the challenges associated with "Celestia-app needs to be careful to register all the appropriate trees once (and only once) before they are used ". What do you see as the main concerns with this requirement?

Nevertheless, I am going to approve and leave the final decision up to you.

staheri14 (approved):

Issue #215, opened 2023-07-05
Consistent naming: shard, chunk, buffer, share


rootulp:
This repo uses the terms: `shard`, `chunk`, `buffer`, and `share` interchangeably. It may help readability if one word was chosen and usage was consistent across the repo.

rootulp:
Update from #299 is to consolidate on share.

musalbas:
This is just my opinion not a decision. Please get inputs from other engineers first such as @adlerjohn @renaynay @Wondertan.

musalbas:
This is also relevant to the docs. Cc @jcstein @YazzyYaz

adlerjohn:
Leaning towards agreeing with the conclusion in #299 to go with `share`.

Issue #262, opened 2023-08-07
Inconsistent terminology in variable names: chunks vs shares


musalbas:
We use chunks and shares interchangeably in variable names and comments - we should stick to one, probably "chunks".

rootulp:
Similar to https://github.com/celestiaorg/rsmt2d/issues/215. 

Out of curiosity why the preference for "chunks"? Since "shares" are commonly used in celestia-app, I have a slight preference for it.

musalbas:
No particular reason other than it seems more intuitive or understandable
due to being less formal.

PR #299, opened 2024-02-21
refactor: rename share to chunk

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/215

I opted to use chunk instead of share because @musalbas preferred it and chunk already appears in the exported API of this library (e.g. `ValidateChunkSize`, `MaxChunks`). 

Technically `Shares` is in the public API too (it's a field of `ErrByzantineData`) but this PR doesn't introduce any breaking changes b/c IMO it's not worth breaking API over this name change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,17 +18,17 @@ import (
 )
 
 func main() {
-    // shareSize is the size of each share (in bytes).
-    shareSize := 512
-    // Init new codec
+    // chunkSize is the size of each chunk in bytes.
+    chunkSize := 512
+    //  new codec
     codec := rsmt2d.NewLeoRSCodec()
 
-    ones := bytes.Repeat([]byte{1}, shareSize)
-    twos := bytes.Repeat([]byte{2}, shareSize)
-    threes := bytes.Repeat([]byte{3}, shareSize)
-    fours := bytes.Repeat([]byte{4}, shareSize)
+    ones := bytes.Repeat([]byte{1}, chunkSize)
+    twos := bytes.Repeat([]byte{2}, chunkSize)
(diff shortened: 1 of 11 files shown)


musalbas:
What about celestia-node, celestia-app, and other repos?
Cc @renaynay @Wondertan

rootulp:
celestia-node has [~183](https://github.com/search?q=repo%3Acelestiaorg%2Fcelestia-node+share&type=code) occurrences of share and [~4](https://github.com/search?q=repo%3Acelestiaorg%2Fcelestia-node+chunk&type=code) for chunk. celestia-app almost always uses shares and not chunk. I had a preference for share but you preferred chunk. See https://github.com/celestiaorg/rsmt2d/issues/262#issuecomment-1668381460.

musalbas:
If share is overwhelmingly used more than chunk then I'm not against using share

musalbas:
But in this repo it seems that chunk is used more than share?

rootulp:
This repo is split. 188 occurrences of chunk and 300 occurrences of share. I still lean slightly towards share because usage in downstream repos but not strongly opinionated on it. Can do whichever reviewers prefer.

musalbas:
Wonder what Celestia node API uses. I also do think it seems people say
share more than chunks

rootulp:
celestia-node API uses share: https://node-rpc-docs.celestia.org/?version=v0.13.0#share. I don't see instances of chunk in those docs.

musalbas:
If most apis and docs use share I'd guess it makes more sense to use share
in the code

rootulp:
Agreed, can do.

PR #306, opened 2024-03-05
refactor: chunk -> share

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/215

### DIFF
--- a/codecs.go
+++ b/codecs.go
@@ -19,12 +19,13 @@ type Codec interface {
 	// Missing shares must be nil. Returns original + parity data.
 	Decode(data [][]byte) ([][]byte, error)
 	// MaxChunks returns the max number of chunks this codec supports in a 2D
-	// original data square.
+	// original data square. Chunk is a synonym of share.
 	MaxChunks() int
 	// Name returns the name of the codec.
 	Name() string
 	// ValidateChunkSize returns an error if this codec does not support
-	// chunkSize. Returns nil if chunkSize is supported.
+	// chunkSize. Returns nil if chunkSize is supported. Chunk is a synonym of
+	// share.
 	ValidateChunkSize(chunkSize int) error
 }
 
--- a/codecs.go
+++ b/codecs.go
@@ -6,8 +6,8 @@ const (
 	// Leopard is a codec that was originally implemented in the C++ library
 	// https://github.com/catid/leopard. rsmt2d uses a Go port of the C++
 	// library in https://github.com/klauspost/reedsolomon. The Leopard codec
-	// uses 8-bit leopard for shards less than or equal to 256. The Leopard
-	// codec uses 16-bit leopard for shards greater than 256.
+	// uses 8-bit leopard for shares less than or equal to 256. The Leopard
+	// codec uses 16-bit leopard for shares greater than 256.
 	Leopard = "Leopard"
 )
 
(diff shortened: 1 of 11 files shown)


codecs.go
```diff
@@ -19,12 +19,13 @@ type Codec interface {
 	// Missing shares must be nil. Returns original + parity data.
 	Decode(data [][]byte) ([][]byte, error)
 	// MaxChunks returns the max number of chunks this codec supports in a 2D
-	// original data square.
+	// original data square. Chunk is a synonym of share.
 	MaxChunks() int
```
staheri14, line 23: [suggestion] How about introducing a new method named `MaxShares()` and advising its use as a preferable alternative to` MaxChunks`? The same for `ValidateChunckSize()` -> `ValidateShareSize()`

datasquare.go
```diff
@@ -9,8 +9,9 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-// ErrUnevenChunks is thrown when non-nil chunks are not all of equal size.
-var ErrUnevenChunks = errors.New("non-nil chunks not all of equal size")
+// ErrUnevenChunks is thrown when non-nil shares are not all of equal size.
+// Note: chunks is synonymous with shares.
+var ErrUnevenChunks = errors.New("non-nil shares not all of equal size")
```
staheri14, line 14: Introducing a new variable for `ErrUnevenShares` could indeed enhance clarity. However, it appears that this PR deliberately avoids making breaking changes, which is why there are still functions, methods, and variables containing `Chunck` in their names. The coexistence of identifiers with both `chunck`s and `share`s might lead to confusion, though.

staheri14 (approved):
My suggestions are optional, the PR as it is LGTM!

codecs.go
rootulp: I thought of this but I think it's breaking to add them to the existing interface

datasquare.go
rootulp: Yep agreed. I'd like to completely eliminate chunk in favor of share but it would be breaking to do so for a few places so I left the existing "chunk" and tried to include comments saying chunk == share

codecs.go
staheri14: It would a breaking change, and if the aim is to escape breaking changes, then the interface should not be touched.

rootulp:
gah tests are failing because https://github.com/celestiaorg/rsmt2d/actions/runs/8207511252/job/22448943530?pr=306#step:6:17

extendeddatasquare.go
```diff
@@ -109,18 +111,18 @@ func ImportExtendedDataSquare(
 // NewExtendedDataSquare returns a new extended data square with a width of
 // edsWidth. All shares are initialized to nil so that the returned extended
 // data square can be populated via subsequent SetCell invocations.
-func NewExtendedDataSquare(codec Codec, treeCreatorFn TreeConstructorFn, edsWidth uint, chunkSize uint) (*ExtendedDataSquare, error) {
+func NewExtendedDataSquare(codec Codec, treeCreatorFn TreeConstructorFn, edsWidth uint, shareSize uint) (*ExtendedDataSquare, error) {
```
adlerjohn, line 114: This looks like it's part of the public API, right? If so this is a breaking change.

staheri14 (approved):

evan-forbes (approved):

rootulp:
Going to merge as is. If we need to be really careful and not introduce any technically API breaking changes, I can revert the few name changes that manifest in the public API docs.

PR #310, opened 2024-04-09
fix: deepcopy filliing nil data with zeroes

walldiss:
Corrects the bug introduced by https://github.com/celestiaorg/rsmt2d/pull/100. Currently, if `deepcopy()` encounters nil inner slice values, it creates a slice of 0 length, which is incorrect. This PR addresses the issue. The problem becomes evident when eds is partially computed, resulting in Row() or Col() methods returning invalid copies.

On the side note, it is technically breaking public API change. But the nature of the bug suggests noone relied on existing misbehaviour before.

### DIFF
--- a/extendeddatasquare.go
+++ b/extendeddatasquare.go
@@ -267,6 +267,9 @@ func (eds *ExtendedDataSquare) RowRoots() ([][]byte, error) {
 func deepCopy(original [][]byte) [][]byte {
 	dest := make([][]byte, len(original))
 	for i, cell := range original {
+		if cell == nil {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


Wondertan:
The rust mind cannot comprehend this :joy:

Wondertan (approved):
Minor correction: The bug was there before the linked PR. The PR extracted existing code as is.

walldiss:
I wish golang had immutables too 🤝 
Updated desc, bug was introduced in previous PR https://github.com/celestiaorg/rsmt2d/pull/100

rootulp (approved):

evan-forbes (approved):

PR #312, opened 2024-04-09
test: add randomized tests for errByzantine

walldiss:
ErrByzantine encompasses numerous edge cases. To address this, I implemented randomized fuzzing tests, which identified instances where shares encoding was not properly verified. 

While I am neutral about maintaining the pretty printing code (as it looks kinda off), I retained it because it proved extremely useful for understanding the reasons behind test failures during extensive debugging. Should these tests fail in the future for any reason, having a visual representation of the issues will expedite debugging for future developers.

Allowed to discover https://github.com/celestiaorg/rsmt2d/pull/313

### DIFF
--- a/extendeddatacrossword_test.go
+++ b/extendeddatacrossword_test.go
@@ -491,9 +491,9 @@ func repairNewFromCorrupted(codec Codec, corrupted *ExtendedDataSquare, corrupte
 	}
 
 	// set corrupted share first
-	errX, errY := corruptedIdx/int(corrupted.Width()), corruptedIdx%int(corrupted.Width())
-	share := corrupted.GetCell(uint(errX), uint(errY))
-	err = square.SetCell(uint(errX), uint(errY), share)
+	corruptedX, corruptedY := corruptedIdx/int(corrupted.Width()), corruptedIdx%int(corrupted.Width())
+	share := corrupted.GetCell(uint(corruptedX), uint(corruptedY))
+	err = square.SetCell(uint(corruptedX), uint(corruptedY), share)
 	if err != nil {
 		return nil, fmt.Errorf("failure to set corrupted share: %w", err)
 	}
@@ -516,7 +516,7 @@ func repairNewFromCorrupted(codec Codec, corrupted *ExtendedDataSquare, corrupte
 		}
 		var errByz *ErrByzantineData
 		if errors.As(err, &errByz) {
-			err = checkErrByzantine(errByz, errX, errY)
+			err = checkErrByzantine(errByz, corruptedX, corruptedY)
 			if err != nil {
 				prettyPrintSamples(samples, corruptedIdx)
 			}


extendeddatacrossword_test.go
```diff
@@ -2,6 +2,7 @@ package rsmt2d
 
 import (
 	"bytes"
+	crand "crypto/rand"
 	"errors"
 	"fmt"
 	"math/rand"
```
rootulp, line 8: [no change needed] in a follow up PR we may consider replacing usage of `math/rand` with `crypto/rand` and then deleting the import alias for `crypto/rand`

extendeddatacrossword_test.go
```diff
@@ -423,6 +424,186 @@ func TestCorruptedEdsReturnsErrByzantineData_UnorderedShares(t *testing.T) {
...
+	t.Skip()
+	for i := 0; i < 10000; i++ {
+		TestErrRandByzantine(t)
+	}
+}
+
+func TestErrRandByzantine(t *testing.T) {
+	codec := NewLeoRSCodec()
+	original, corrupted, idx := randCorruptedEDS(t, codec, 8)
+	require.False(t, original.Equals(corrupted), "corrupted eds is equal to original eds")
```
rootulp, line 438: [question] why `.False` instead of:

```go
	require.NotEqual(t, original, corrupted)
```

extendeddatacrossword_test.go
```diff
@@ -423,6 +424,186 @@ func TestCorruptedEdsReturnsErrByzantineData_UnorderedShares(t *testing.T) {
...
+		shareSize,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failure to create extended data square: %w", err)
+	}
+
+	// set corrupted share first
+	errX, errY := corruptedIdx/int(corrupted.Width()), corruptedIdx%int(corrupted.Width())
+	share := corrupted.GetCell(uint(errX), uint(errY))
+	err = square.SetCell(uint(errX), uint(errY), share)
```
rootulp, line 496: [nit] perhaps avoid `errX` and `errY` because it's easy to visually conflate them with `err`. Candidates: `corruptedX` and `corruptedY`?

extendeddatacrossword_test.go
```diff
@@ -423,6 +424,186 @@ func TestCorruptedEdsReturnsErrByzantineData_UnorderedShares(t *testing.T) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failure to get column roots: %w", err)
+	}
+
+	// loop until repaired or byzantine error
+	for {
+		repaired, err := fillRandomCellAndRepair(corrupted, square, rowRoots, colRoots, samples)
+		if repaired {
+			prettyPrintSamples(samples, corruptedIdx)
+			return square, errors.New("no byzantine error")
```
rootulp, line 515: [question] why return an error her? Why not

```suggestion
			return square, nil
```

rootulp (approved):
LGTM

extendeddatacrossword_test.go
walldiss: Test corrupts eds and tries to repair it. It shoud result in Byzantine every time on corrupted eds. If there was no encoding errors during eds construction, it means we missed some cases.

extendeddatacrossword_test.go
walldiss: sgtm https://github.com/celestiaorg/rsmt2d/pull/312/commits/61f82d1333a6bd5d3e483b56f108ba00d70f1c16

extendeddatacrossword_test.go
walldiss: Unfortunately, `crypto/rand` lacks handy IntN funciton, only 
```
func Int(rand io.Reader, max *big.Int) (n *big.Int, err error) 
```
While `math/rand` lacks ability to generate random []byte slice. I wish one day there will be single package for both

extendeddatacrossword_test.go
walldiss: NotEqual uses `==` slice compare under the hood, while `Equals` method does more precise comparison.

extendeddatacrossword_test.go
Wondertan: There is rand v2

Wondertan (approved):

evan-forbes (approved):

PR #313, opened 2024-04-09
fix: verify encoding of shares in case when axis is fully reconstructed from orthogonal

walldiss:
If axis is fully recomputed from orthogonal axises, we need to verify correct encoding on top of verification of roots.
Test coverage should happen by https://github.com/celestiaorg/rsmt2d/pull/312

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -471,6 +471,7 @@ func (eds *ExtendedDataSquare) computeSharesRootWithRebuiltShare(shares [][]byte
 	return tree.Root()
 }
 
+// verifyEncoding checks the Reed-Solomon encoding of the provided data.
 func (eds *ExtendedDataSquare) verifyEncoding(data [][]byte, rebuiltIndex int, rebuiltShare []byte) error {
 	data[rebuiltIndex] = rebuiltShare
 	half := len(data) / 2


extendeddatacrossword.go
```diff
@@ -138,10 +138,7 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 
 	// Prepare shares
 	shares := make([][]byte, eds.width)
-	vectorData := eds.row(uint(r))
-	for c := 0; c < int(eds.width); c++ {
-		shares[c] = vectorData[c]
-	}
+	copy(shares, eds.row(uint(r)))
```
walldiss, line 141: just small nicetohave change to use builtin func instead of loop

Wondertan:
Do you know why codecov failed? 🤔

Wondertan:
This seems like an important case we missed before and needs rsmt2d OGs to review it
сс @adlerjohn and @musalbas

adlerjohn (approved):
Cute edge case, good catch.

rootulp (approved):

staheri14 (approved):

extendeddatacrossword.go
```diff
@@ -468,3 +470,22 @@ func (eds *ExtendedDataSquare) computeSharesRootWithRebuiltShare(shares [][]byte
 	}
 	return tree.Root()
 }
+
+func (eds *ExtendedDataSquare) verifyEncoding(data [][]byte, rebuiltIndex int, rebuiltShare []byte) error {
```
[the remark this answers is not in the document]
walldiss: done

rootulp (approved):

extendeddatacrossword.go
```diff
@@ -177,6 +174,10 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 				}
 				return false, false, err
 			}
+
+			if eds.verifyEncoding(col, r, rebuiltShares[c]) != nil {
```
staheri14, line 178: Something worth consideration is the the precedence of these returned errors i.e., 1) root mismatch (due to the call to `verifyAgainstColRoots`) and 2) the encoding error (caused by `verifyEncoding`). I think if the encoding is already wrong, perhaps it does not matter whether the roots match or not, so returning the encoding verification error would be more accurate and informative. Depending on this decision, we may want to swap the order of the returned errors.  This can be tackled in a separate issue though if further discussion is needed.

staheri14 (approved):
LGTM!

PR #318, opened 2024-04-19
fix: verifyEncoding should revert changes in verifiable data

walldiss:
The VerifyEncoding function introduced in PR #313 might not revert changes to the original row or column after successful verification. This could lead to a scenario where one of the fields, either squareRow or squareCol, retains a rebuilt share, resulting in only partial repair. This oversight would leave some repaired data unavailable when it should be accessible. Existing tests did not cover this scenario. This PR addresses the issue and includes randomized repair tests to ensure the fix is effective.

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -378,11 +378,8 @@ func (eds *ExtendedDataSquare) preRepairSanityCheck(
 				return nil
 			})
 			errs.Go(func() error {
-				parityShares, err := eds.codec.Encode(eds.rowSlice(i, 0, eds.originalDataWidth))
+				err := eds.verifyEncoding(eds.row(i), noShareInsertion, nil)
 				if err != nil {
-					return err
-				}
-				if !bytes.Equal(flattenShares(parityShares), flattenShares(eds.rowSlice(i, eds.originalDataWidth, eds.originalDataWidth))) {
 					return &ErrByzantineData{Row, i, eds.row(i)}
 				}
 				return nil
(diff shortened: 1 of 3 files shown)


rootulp (approved):

Wondertan (approved):
utACK

extendeddatacrossword.go
```diff
@@ -473,7 +466,14 @@ func (eds *ExtendedDataSquare) computeSharesRootWithRebuiltShare(shares [][]byte
 
 // verifyEncoding checks the Reed-Solomon encoding of the provided data.
 func (eds *ExtendedDataSquare) verifyEncoding(data [][]byte, rebuiltIndex int, rebuiltShare []byte) error {
-	data[rebuiltIndex] = rebuiltShare
+	if rebuiltShare != nil && rebuiltIndex >= 0 {
+		data[rebuiltIndex] = rebuiltShare
+		defer func() {
```
adlerjohn, line 471: Style: do we want to use `defer` here? AFAIK, no where else in rsmt2d uses `defer`, primarily because it's a Go-specific thing that doesn't translate well into a language like C (C being a lower common denominator for readability). I don't have strong opinions, so opening for discussion.

walldiss: I'm in favor of using defer because it ensures cleanup operations are closely tied to their corresponding setups. This not only makes the code more robust by preventing omission in multiple return paths but also aligns with good programming practices by keeping related actions together, improving both maintainability and readability.

An alternative would be to manually add a revert of the addition to the data slice in every return statement, which could be error-prone. 

Another option is to adopt an immutable approach by creating a copy of the data that includes the extra share, ensuring the original data remains unmodified. However, this method would lead to extra allocations, which is why I haven't used it here.

adlerjohn: I see.

Hmmmm. Perhaps creating a copy wouldn't be so bad if it's only for a single row or column. Could you open an issue that someone can work on to investigate the actual performance impact?

Other that than, no further comments. Not blocking merging.

Wondertan: Defer translates well into C or any other language by calling the deferred line after everything else, so we don't gain much by going with the alternative approach and only lose time (investigating performance impact through the issue) and readability(no defer based grouping). Overall, I don't think opening and spending time on the issue is worth it

adlerjohn: Ok. I'll defer (get it 😉) to you guys on this.

Issue #350, opened 2025-04-28
ExtendedDataSquare.solveCrosswordRow captures an ErrByzantineData error but discards it inexplicably: code looks incomplete


odeke-em:
https://github.com/celestiaorg/rsmt2d/blob/a3f832ad21f46d55d9390b4110e8ef03c4375a16/extendeddatacrossword.go#L155-L159

and also

https://github.com/celestiaorg/rsmt2d/blob/a3f832ad21f46d55d9390b4110e8ef03c4375a16/extendeddatacrossword.go#L229-L233
look quite incomplete. What was the point of the `errors.As(err, &byzErr)` if nothing to do with it? That code looks incomplete and we need to figure out what to do with that error and how to return it.

rootulp:
The error is returned in the code snippet you shared. Specifically https://github.com/celestiaorg/rsmt2d/blob/a3f832ad21f46d55d9390b4110e8ef03c4375a16/extendeddatacrossword.go#L159 and https://github.com/celestiaorg/rsmt2d/blob/a3f832ad21f46d55d9390b4110e8ef03c4375a16/extendeddatacrossword.go#L233

odeke-em:
Sure, you are correct but it is a use of side effects but, that's definitely a non-clear way of using errors.As and contrary to the point of using errors.As which is that you retrieve the underlying error and then use the target value as is. Kindly please make it clear in code to return that value.

PR #351, opened 2025-04-28
refactor: return `byzErr` for clarity

rootulp:
Closes https://github.com/celestiaorg/rsmt2d/issues/350

This code change is strictly for clarity and it should behave identically to before. These tests still pass:
1. https://github.com/rootulp/rsmt2d/blob/fe785dafe00a1cc6cbdd346be1bd01fd193f758e/extendeddatacrossword_test.go#L185
2. https://github.com/rootulp/rsmt2d/blob/fe785dafe00a1cc6cbdd346be1bd01fd193f758e/extendeddatacrossword_test.go#L346

cc: @odeke-em

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -155,6 +155,7 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 		var byzErr *ErrByzantineData
 		if errors.As(err, &byzErr) {
 			byzErr.Shares = shares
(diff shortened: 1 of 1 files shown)


evan-forbes (approved):

Wondertan (approved):

adlerjohn:
Is this actually functionally identical on the error output? The original code looks like it returns a strictly different error.

PR #395, opened 2026-02-09
fix missing share randomization in decoding benchmark

Ocheretovich:
Fix decoding benchmark to randomize missing shares across all shares, not only data


rootulp (approved):

rootulp:
Ran `BenchmarkDecoding` locally with this change applied on darwin/arm64 (Apple M4 Pro, Go test `-benchtime=3x`):

```
BenchmarkDecoding/Leopard_128_shares_512-12    3    2236 ns/op
PASS
```

Benchmark runs cleanly — `reedsolomon.Reconstruct` handles the `nil` entries correctly (that's the library's convention for missing shares), and the change also aligns with how the rest of the codebase treats missing shares (e.g. `extendeddatacrossword.go`, `datasquare.go`). LGTM.

PR #411, opened 2026-05-22
fix: attach correct-axis, nil-free shares to ErrByzantineData

evan-forbes:
Fixes two related issues in the `ErrByzantineData.Shares` evidence returned by `Repair`:

- **Wrong-axis / nil-hole evidence:** the orthogonal-completion loops in `solveCrosswordRow`/`solveCrosswordCol` attached the rebuilt other-axis buffer to an error that names the orthogonal axis; they now attach a copy of the named axis with the rebuilt share inserted (and the same-axis returns snapshot committed shares before in-place decoding to preserve nils).
- **`verifyEncoding` mutating the caller slice:** it wrote the rebuilt share into the caller-owned row/column slice and deferred a reset to nil, leaving a nil hole in the attached shares; it now uses a defensive copy and never mutates its input.

These `Repair`/BEFP code paths are not reachable so this is strictly hardening. Bad-encoding fraud proofs and reconstruction are no longer enabled and not planned to be maintained. This is fixed simply because it is cheap, correct hardening.

### DIFF
--- a/extendeddatacrossword.go
+++ b/extendeddatacrossword.go
@@ -139,6 +139,9 @@ func (eds *ExtendedDataSquare) solveCrosswordRow(
 	// Prepare shares
 	shares := make([][]byte, eds.width)
 	copy(shares, eds.row(uint(rowIdx)))
+	// Snapshot committed shares (nils preserved) before rebuildShares decodes in place.
+	committed := make([][]byte, eds.width)
+	copy(committed, shares)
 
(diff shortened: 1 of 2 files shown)


evan-forbes:
converting to a draft while I take a deeper look. I not actually sure we want to patch this

rootulp:
We should close in favor of https://github.com/celestiaorg/rsmt2d/pull/417

PR #52, opened 2021-04-20
Optimization: store duplicate rows and columns

adlerjohn:
Implements the second choice of optimization in https://github.com/lazyledger/rsmt2d/issues/39#issuecomment-817107324. Refactoring everything to go through `Call(row, col)` would be an astronomical amount of work, as e.g. encoding libraries expect simply `[][]byte` slices. This implementation increases memory footprint by ~2 for the data square, but non-trivially reduces runtime especially at larger square sizes, as returning slices is zero-allocation.

The `Repair` benchmarks below are most representative of real-world use, and see _increasing_ decreases in runtime as square size increases.

<details>
<summary>Click to expand</summary>

```
name                                        old time/op    new time/op    delta
Encoding/Encoding_128_shares_using_RSGF8-8     144µs ± 0%     145µs ± 0%   +0.15%  (p=1.000 n=1+1)
Decoding/Decoding_128_shares_using_RSGF8-8    30.1µs ± 0%    30.6µs ± 0%   +1.74%  (p=1.000 n=1+1)
Roots/Square_Size_32x32-8                      603ns ± 0%     612ns ± 0%   +1.46%  (p=1.000 n=1+1)
Roots/Square_Size_64x64-8                     1.31µs ± 0%    1.46µs ± 0%  +11.04%  (p=1.000 n=1+1)
Roots/Square_Size_128x128-8                   2.08µs ± 0%    2.20µs ± 0%   +5.87%  (p=1.000 n=1+1)
Roots/Square_Size_256x256-8                   3.56µs ± 0%    3.79µs ± 0%   +6.64%  (p=1.000 n=1+1)
Repair/Repairing_16x16_ODS_using_RSGF8-8      3.60ms ± 0%    3.32ms ± 0%   -7.78%  (p=1.000 n=1+1)
Repair/Repairing_32x32_ODS_using_RSGF8-8      17.4ms ± 0%    14.8ms ± 0%  -15.04%  (p=1.000 n=1+1)
Repair/Repairing_64x64_ODS_using_RSGF8-8      80.0ms ± 0%    63.6ms ± 0%  -20.45%  (p=1.000 n=1+1)
Repair/Repairing_128x128_ODS_using_RSGF8-8     419ms ± 0%     312ms ± 0%  -25.70%  (p=1.000 n=1+1)
Extension/RSGF8_size_4-8                      12.1µs ± 0%    12.7µs ± 0%   +4.94%  (p=1.000 n=1+1)
[...]

PR #412, opened 2026-06-01
docs: fix incorrect comment on dataSquare

### LINKED ISSUE #280: Fix incorrect commen on `datasquare`
The comment [states](https://github.com/celestiaorg/rsmt2d/blob/2f682ad7bacfcf738841fa0df8230ac6ffebacc9/datasquare.go#L16-L17) that data is duplicated in the square, but it's not. It would copy if the [line](https://github.com/celestiaorg/rsmt2d/blob/2f682ad7bacfcf738841fa0df8230ac6ffebacc9/datasquare.go#L59) did a `copy` rather than assigning, which simply only copies pointers to slices, rather than actual slices.

rootulp:
## Overview

Fixes the misleading comment on the `dataSquare` type. The comment claimed that share data is "duplicated in both row-major and column-major order", but this is not what the code does.

In `newDataSquare`:
- `squareRow[rowIdx] = data[rowIdx*width : rowIdx*width+width]` stores sub-slices of `data`.
- `squareCol[colIdx][rowIdx] = data[rowIdx*width+colIdx]` stores the same `[]byte` elements.

So `squareRow[rowIdx][colIdx]` and `squareCol[colIdx][rowIdx]` both hold the **same `[]byte` slice header**, pointing to the same backing array. Only the pointers/references are arranged in two layouts; the underlying share bytes are **not** duplicated. `SetCell`, `setRowSlice`, and `setColSlice` confirm this by writing the identical pointer into both layouts.

As noted in the issue, [PR #52](https://github.com/celestiaorg/rsmt2d/pull/52) was supposed to introduce the duplication but assigned slices rather than copying bytes, so the duplication never happened; the benchmark improvements there came from reduced allocations elsewhere.

This PR updates the comment to accurately describe the data structure.

Closes https://github.com/celestiaorg/rsmt2d/issues/280

### DIFF
--- a/datasquare.go
+++ b/datasquare.go
@@ -14,8 +14,10 @@ import (
 var ErrUnevenChunks = errors.New("non-nil shares not all of equal size")
 
 // dataSquare stores all data for an original data square (ODS) or extended
-// data square (EDS). Data is duplicated in both row-major and column-major
-// order in order to be able to provide zero-allocation column slices.
+// data square (EDS). Pointers to the shares are stored in both row-major
+// (squareRow) and column-major (squareCol) order. Both layouts reference the
+// same underlying share byte slices (the share bytes themselves are not
+// duplicated), which enables zero-allocation row and column slices.
 type dataSquare struct {
 	squareRow    [][][]byte // row-major
 	squareCol    [][][]byte // col-major


ninabarbakadze (approved):
lgtm. somehow missed this.

PR #418, opened 2026-06-11
fix: UnmarshalJSON limited to the default tree

### LINKED ISSUE #275: `UnmarshalJSON` is limited to the default Tree

rootulp:
## Overview

Closes https://github.com/celestiaorg/rsmt2d/issues/275 (no associated Linear issue).

`ExtendedDataSquare.UnmarshalJSON` hardcoded `NewDefaultTree`, so an EDS built with a custom tree (i.e. Celestia's erasured NMT — the only tree production software ever serializes) round-tripped through JSON into an EDS whose recomputed row/col roots were wrong against any real DAH.

This is the minimal fix: `UnmarshalJSON` now reconstructs the square with the erasured NMT tree instead of the default tree. Unlike the previous registry-based fix (#277/#287, reverted in #295), there is no registry, no new exported API, and no wire format change — nothing for celestia-app/node to do besides bump the dependency.

- Promotes the erasured NMT wrapper (previously test-only, adapted from celestia-app) into production code. The NMT options are hardcoded to Celestia's: sha256, namespace size 29, `ignoreMaxNamespace=true`.
- `UnmarshalJSON` derives the original square width from the data itself and reconstructs with the NMT tree.
- `UnmarshalJSON` now returns errors for an unsupported codec name (previously panicked on a nil codec) and for an empty `data_square` (previously produced a degenerate width-0 EDS).
- Fixes a latent nil dereference on the error path of `deepCopy`.

## Compat note (deliberate behavior change)

Deserialized squares now compute NMT roots instead of default-tree roots. Every EDS ever serialized by production Celestia software is an NMT-built square, so this makes all pre-existing JSON round-trip with correct roots — celestia-node is fixed by bumping rsmt2d, with no downstream code change. The old default-tree reconstruction only produced roots that were wrong against any real DAH, so nothing can depend on it; the shares are identical either way. A code search found no production use of the default tree anywhere; it remains in the codebase as a test fixture. Squares built with non-NMT trees do not round-trip their roots through JSON (documented in the README).

## API surface

No new exported symbols. No existing signatures changed. Wire format unchanged.

## Testing

- Regression test `TestUnmarshalJSONNMTRoots`: an NMT-built EDS (mimicking exactly what celestia-node serializes today) → marshal → unmarshal → identical row/col roots.
- `TestUnmarshalJSONErrors`: unsupported codec name, empty data square.
- `go test -race ./...` and `golangci-lint run` pass.


tree.go
```diff
@@ -57,3 +58,27 @@ func (d *DefaultTree) Root() ([]byte, error) {
...
+// treeConstructorByName returns the TreeConstructorFn for one of the built-in
+// tree names. odsWidth is the width of the original (unextended) data square;
+// it is required by trees whose shape depends on the square size and ignored
+// by the others.
+func treeConstructorByName(treeName string, odsWidth uint) (TreeConstructorFn, error) {
+	switch treeName {
+	case DefaultTreeName:
+		return NewDefaultTree, nil
+	case NMTTreeName:
+		return nmtTreeConstructor(odsWidth), nil
```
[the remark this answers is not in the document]
rootulp: Added the guard in 4a692fb, with one correction to the rationale: the 1-element-slice scenario described here does not panic on this branch. `newDataSquare` only stores the constructor (trees are built lazily at root-computation time), so `validateEdsWidth` rejects the odd width first — `ImportExtendedDataSquareWithTree([][]byte{share}, codec, NMTTreeName)` returns `extended data square width 1 must be even`, verified by test.

The guard is still worthwhile as fail-fast hardening: an *empty* `data_square` previously round-tripped into a degenerate width-0 EDS with no error, and it keeps the lazily-panicking NMT constructor from ever being created with squareSize 0. That now errors upfront, consistent with the unsupported tree/codec name guards.

tree_test.go
```diff
@@ -0,0 +1,64 @@
...
+		require.NoError(t, err)
+
+		assert.Equal(t, wantRoot, gotRoot)
+	})
+
+	t.Run("returns an error for an unknown tree name", func(t *testing.T) {
+		_, err := treeConstructorByName("unknown-tree", 2)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "unsupported tree name")
+	})
```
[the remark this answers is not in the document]
rootulp: Added in 4a692fb as `returns an error for NMTTreeName with zero odsWidth` (written red-first: it failed before the guard, passes after).

rootulp:
Scope reduction in 445cf0f, requested by @rootulp: the `tree` JSON field, the exported `DefaultTreeName`/`NMTTreeName` constants, and the `WithTree` constructors are removed. `UnmarshalJSON` now simply always reconstructs with the erasured NMT tree — the only tree production Celestia software serializes. Net diff vs main is +169/−5 with zero new exported API and no wire format change.

The zero-odsWidth guard from the earlier review thread now lives in `UnmarshalJSON` (empty `data_square` → error) since `treeConstructorByName` no longer exists. PR description has been rewritten to match.

tree.go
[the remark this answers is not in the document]
rootulp: Final state after the scope reduction in 445cf0f: `treeConstructorByName` no longer exists; the equivalent guard lives in `UnmarshalJSON` (empty `data_square` → `data square must not be empty`), so a zero square size can never reach `nmtTreeConstructor` from the JSON path. Resolving.

tree_test.go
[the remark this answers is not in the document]
rootulp: Covered by `TestUnmarshalJSONErrors/returns an error for an empty data square` after the scope reduction in 445cf0f (`tree_test.go` was removed along with `treeConstructorByName`). Resolving.

rootulp:
I can't request an explicit review from @Wondertan but tagging him for awareness. This approach diverges from the proposal in the issue but it's the minimal diff that would resolve that issue. I expect @Wondertan to push back so his PR feedback may help

nmtwrapper.go
```diff
@@ -142,3 +144,19 @@ func (w *erasuredNamespacedMerkleTree) incrementShareIndex() {
...
+
+// nmtTreeConstructor returns the TreeConstructorFn for Celestia's erasured
+// namespaced Merkle tree, for a square whose original (unextended) width is
+// odsWidth. The underlying NMT uses sha256, a namespace size of
+// celestiaNamespaceSize, and ignoreMaxNamespace=true, matching celestia-app's
+// wrapper.ErasuredNamespacedMerkleTree.
+func nmtTreeConstructor(odsWidth uint) TreeConstructorFn {
+	return newErasuredNamespacedMerkleTreeConstructor(uint64(odsWidth),
+		nmt.NamespaceIDSize(celestiaNamespaceSize))
+}
```
vgonkivs, line 162: This bakes Celestia-specific knowledge directly into rsmt2d's core, which I think inverts the dependency direction the library is built around. rsmt2d is meant to be tree-agnostic — the whole point of TreeConstructorFn /NewDefaultTree is that the tree is pluggable and the library doesn't know about  any particular consumer. Hardcoding the erasured NMT (plus celestiaNamespaceSize = 29, sha256, ignoreMaxNamespace) in UnmarshalJSON makes a generic  library depend on its specific consumer, and promotes a copy of celestia app's wrapper.ErasuredNamespacedMerkleTree into rsmt2d production code — now  two copies to keep in sync.

  A few concrete concerns:
  - Magic constant 29: this is Celestia's namespace size (1 version + 28 ID). If app-side namespacing ever changes, JSON round-trips here break silently, with the divergence buried in a third-party lib rather than surfaced in app code.
  - Layering: the dependency should flow app → rsmt2d, not rsmt2d → app concepts.

  I understand the motivation — the JSON doesn't carry which TreeConstructorFn was used, and NewDefaultTree produces wrong roots for real squares.
 But rather than hardcoding the one tree we happen to serialize today, could we keep rsmt2d generic and let the caller supply the constructor? e.g.:
  - an UnmarshalJSON variant / option that accepts a TreeConstructorFn, or a settable package-level default, so celestia-app/node injects the erasured
  NMT, or
  - serialize the tree identity in the JSON so the round-trip reconstructs correctly without out-of-band assumptions, or
  - keep the "JSON ↔ erasured NMT" logic in celestia-app/celestia-node, where the NMT knowledge already lives.

rootulp: This library doesn't need to be generic and it isn't used by anyone besides Celestia. I tried handling this while remaining generic and it involves a much larger code change.

vgonkivs: I'd frame this as a change to the dependency contract. Today it's one-directional:
*   app  ->  rsmt2d
  app imports rsmt2d; rsmt2d knows nothing about app. This PR makes it bidirectional:
  
*   app  <-> rsmt2d
rsmt2d now encodes knowledge that app owns — the erasured NMT and celestiaNamespaceSize = 29. It isn't a compile-time  import cycle (rsmt2d still doesn't import app), which is exactly what makes it worse: a real back-edge would force a single source of truth, but a  copied constant + wrapper is an unenforced back-edge that drifts silently. The day app changes its namespace size, nothing — not the compiler, not a test — connects the two, and the JSON path diverges inside a dependency.

And if keeping the edge one-directional needs more code, I'd rather take that larger refactor. Diff size shouldn't decide the dependency direction.

> This library doesn't need to be generic and it isn't used by anyone besides Celestia

Fair enough on the generic point.

vgonkivs: To make this concrete: a MustRegister at startup resolves it cleanly and keeps the arrow one-directional. The tree lives in rsmt2d's global state (same init-time pattern as prometheus.MustRegister / the bitswap block registry), the app wires its NMT once, and if nothing is registered we fail loud instead of silently reconstructing with the wrong tree:

```go
  // rsmt2d knows nothing about namespaces, only that a tree was provided.
  var unmarshalTreeFor func(odsWidth uint) TreeConstructorFn

  // MustRegisterUnmarshalTree sets the tree UnmarshalJSON uses to reconstruct
  // squares. Call once at startup. Registering nil, or twice, panics.
  func MustRegisterUnmarshalTree(f func(odsWidth uint) TreeConstructorFn) {
        if f == nil {
                panic("rsmt2d: MustRegisterUnmarshalTree: nil constructor")
        }
        if unmarshalTreeFor != nil {
                panic("rsmt2d: MustRegisterUnmarshalTree: tree already registered")
        }
        unmarshalTreeFor = f
  }
```

rootulp: I dislike the start up registration pattern. It adds complexity that I don't think is necessary. At a high level, this library doesn't need to remain generic. 

I'll close this PR as won't do. Given we disagree about the correct approach to fix the issue, one alternative is to move the contents of this repo entirely into the celestia-app.

PR #442, opened 2026-09-21
ci: add codecov.yml with 1% project coverage threshold

rootulp:
Follow-up to https://github.com/celestiaorg/rsmt2d/pull/440, where `codecov/project` failed with a 0.69% drop even though every new line was covered and `codecov/patch` passed.

With no `codecov.yml` the project status uses Codecov's default threshold of 0, so any drop below the base commit fails. Total coverage on main swings by close to 1% between commits with identical Go source, because several tests pick which share to corrupt with unseeded `math/rand` and so exercise different repair branches each run. This adds a `codecov.yml` giving the project status a 1% threshold to absorb that noise while still catching real regressions, and keeps the patch status strict. Making those tests deterministic is a separate follow-up.

Verified with Codecov's validator (`curl --data-binary @codecov.yml https://codecov.io/validate` returns Valid).


rootulp:
Superseded by #446, which landed a `codecov.yml` covering the same intent (2% project threshold plus default-branch fix), and #447 removed the underlying test nondeterminism. After merging main this PR's diff is empty, so closing.

Issue #58, opened 2021-05-05
Increase test coverage


adlerjohn:
1. Increase test coverage.
2. Use `rsmt2d_test` package to test all exported methods.
3. Consider using `rsmt2d_test` test package for more of the tests, to not rely on internal methods as much.

PR #447, opened 2026-09-21
test: deflake byzantine coverage and remove unreachable uneven-share check

### LINKED ISSUE #444: Deflake nondeterministic coverage in two byzantine tests and remove unreachable check
Total statement coverage swings between 89.2% and 91.2% across identical local runs, and Codecov's `main` has bounced between 80.89% and 85.89% with no test changes. Two tests cause this:

- `TestValidFraudProof`: `preRepairSanityCheck` runs row and column checks in an errgroup, so whichever goroutine errors first decides whether the byzantine axis is `Row` or `Col`. The test always compares against the row root, so on `Col` runs the fraud-proof body (and `flattenShares`) is skipped entirely. The test should use the root for the reported axis.
- `TestErrRandByzantine`: corrupts a random cell, so the column-byzantine branch in `solveCrosswordCol` is only exercised some of the time. A deterministic case that corrupts a cell reachable only via the column path should sit alongside the random one.

Also, the inner `ErrUnevenChunks` check inside the row loop of `newDataSquare` (`datasquare.go`) is unreachable: the loop above it already rejects every uneven share. It can never be covered and should be removed.

Related to #58.

rootulp:
Closes #444. Related to #58.

Coverage swung by about 2% between identical runs because two byzantine tests were nondeterministic. `TestValidFraudProof` always compared against the row root, but `preRepairSanityCheck` reports Row or Col depending on which goroutine errors first, so on Col runs the erasure-coding check was silently skipped; it now uses the root for the reported axis and asserts they match. `TestErrRandByzantine` only sometimes reached the column-solve byzantine branch, so a deterministic transpose of `TestErrByzantineDataSharesPreserveNils` now covers it every run. The per-cell `ErrUnevenChunks` check inside `newDataSquare`'s row loop was unreachable and is removed, with `TestInvalidDataSquareCreation` now asserting that error for several share positions.

Full suite passes with `-race`, lint is clean, and the total is stable across repeated runs.


datasquare.go
```diff
@@ -55,12 +55,6 @@ func newDataSquare(data [][]byte, treeCreator TreeConstructorFn, shareSize uint)
 	squareRow := make([][][]byte, width)
 	for rowIdx := 0; rowIdx < width; rowIdx++ {
 		squareRow[rowIdx] = data[rowIdx*width : rowIdx*width+width]
-
-		for colIdx := 0; colIdx < width; colIdx++ {
-			if squareRow[rowIdx][colIdx] != nil && len(squareRow[rowIdx][colIdx]) != int(shareSize) {
-				return nil, ErrUnevenChunks
-			}
-		}
```
rootulp, line 63: This can never return an ERrUnevenChunks because of the check above

```
	for _, d := range data {
		if d != nil && len(d) != int(shareSize) {
			return nil, ErrUnevenChunks
		}
	}
```

so this check is redundant and can be removed

rach-id (approved):

